PR #3, opened 2021-02-16
Fix ARM unaligned access and Android CHIPTool crash

Damian-Nordic:
* Fix a bug in CHIP which may lead to executing mbedTLS code on unaligned data structures.
* Fix a bug in Android CHIPTool which causes a crash when trying to commission a device for the second time.

### DIFF
--- a/src/lib/support/SafePointerCast.h
+++ b/src/lib/support/SafePointerCast.h
@@ -1,0 +1,41 @@
+/*
+ *
+ *    Copyright (c) 2021 Project CHIP Authors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


LuDuda (approved):

src/lib/support/SafePointerCast.h
```diff
@@ -0,0 +1,41 @@
+/*
+ *
+ *    Copyright (c) 2020 Project CHIP Authors
```
kkasperczyk-no, line 3: Fix date to 2021?

kkasperczyk-no (approved):

src/lib/support/SafePointerCast.h
Damian-Nordic: Ah, yes. Good that CHIP is going to remove the dates. It's super annoying :)

PR #9, opened 2021-06-01
[nrf_noup]: Fix nrf_security includes for matter

frkv:
-IF CONFIG_NRF_SECURITY_BACKEND - Add includes from mbedtls_external

Signed-off-by: Frank Audun Kvamtrø <frank.kvamtro@nordicsemi.no>

### DIFF
--- a/config/nrfconnect/chip-module/CMakeLists.txt
+++ b/config/nrfconnect/chip-module/CMakeLists.txt
@@ -99,10 +99,20 @@ if (CONFIG_POSIX_API)
 endif()
 
 if (CONFIG_NORDIC_SECURITY_BACKEND)
+  # This is a temphack to ensure that Matter works with nRF Security refactored code
+  if(CONFIG_NORDIC_SECURITY_PROMPTLESS)
+    zephyr_include_directories($<TARGET_PROPERTY:mbedtls_external,INTERFACE_INCLUDE_DIRECTORIES>)
+    zephyr_include_directories($<TARGET_PROPERTY:mbedcrypto_includes,INTERFACE_INCLUDE_DIRECTORIES>)
+    if(TARGET platform_cc3xx)
+      zephyr_include_directories($<TARGET_PROPERTY:platform_cc3xx,INTERFACE_INCLUDE_DIRECTORIES>)
+    endif()
+    list(APPEND CHIP_CFLAGS -DMBEDTLS_CONFIG_FILE=<nrf-config.h>)
+  else()
     # TODO: Remove when the missing dependency is fixed in nRF Connect SDK
     zephyr_include_directories(${ZEPHYR_BASE}/../nrfxlib/crypto/nrf_cc310_platform/include)
 
     list(APPEND CHIP_CFLAGS -DMBEDTLS_CONFIG_FILE=<nrf-config.h>)
+  endif()
 endif()
 
 zephyr_get_compile_flags(CHIP_CFLAGS_C C)


ioannisg (approved):
Looks fine to me but would like @tejlmand to take a quick look.

ioannisg:
BTW @frkv I noticed a compilation error around including a header file, that looks related to the change.

Damian-Nordic:
> BTW @frkv I noticed a compilation error around including a header file, that looks related to the change.

Yes, the `${ZEPHYR_BASE}/../nrfxlib/crypto/nrf_cc310_platform/include` path was added exactly to address this error. I'm not sure why the path is not added to the include dirs.

config/nrfconnect/chip-module/CMakeLists.txt
```diff
@@ -99,9 +99,7 @@ if (CONFIG_POSIX_API)
 endif()
 
 if (CONFIG_NORDIC_SECURITY_BACKEND)
-    # TODO: Remove when the missing dependency is fixed in nRF Connect SDK
-    zephyr_include_directories(${ZEPHYR_BASE}/../nrfxlib/crypto/nrf_cc310_platform/include)
-
+    zephyr_include_directories($<TARGET_PROPERTY:mbedtls_external,INTERFACE_INCLUDE_DIRECTORIES>)
```
tejlmand, line 102: Isn't Zephyr already linking with `mbedtls_external` ?

Like a: `zephyr_link_libraries(mbedtls_external)`.
If that is the case, then this shouldn't be needed and we should find out why it's not working as expected.

If there is no `zephyr_link_libraries(mbedtls_external)` anywhere then this code looks good.

tejlmand (commented):
Just a question

config/nrfconnect/chip-module/CMakeLists.txt
tejlmand: Note, Zephyr is not linking to mbedtls_external as both `MBEDTLS_BUILTIN=n` and `MBEDTLS_LIBRARY=n` when using Nordic security backend.

tejlmand (approved):

ioannisg:
@Damian-Nordic @frkv are you guys looking into the build failure?

tejlmand:
> @Damian-Nordic @frkv are you guys looking into the build failure?

I notice the sdk-nrf revision is completely off, from the log:
```
Executing in build environment: python3 scripts/setup/nrfconnect/update_ncs.py --update
From https://github.com/nrfconnect/sdk-nrf
.....
Previous HEAD position was 8e8c60897 manifest: Update manifest file with v1.5.0 tags
HEAD is now at 4e3dac58c manifest: Update sdk-zephyr revision
```
which is this revision: https://github.com/nrfconnect/sdk-nrf/commit/4e3dac58c45f649df070ace1c129b793e659f306

and thus everything else is off, like the sdk-zephyr and sdk-nrfxlib revisions which are thus becoming:
https://github.com/nrfconnect/sdk-zephyr/commit/a9d3aef6542445d553614bd723e75e5ee9fb4e45
https://github.com/nrfconnect/sdk-nrfxlib/commit/fb936182435a1ae90fafbb8840e70a45d6f99a4e
```
=== updating nrfxlib (nrfxlib):
--- nrfxlib: fetching, need revision fb936182435a1ae90fafbb8840e70a45d6f99a4e
....
HEAD is now at fb93618 softdevice_controller: rev cf113d186d33fbcbd82300ab74a5516fc069f5b2
```

so of course nothing will work, as all the required TF-M integration work is not included when testing this PR.

config/nrfconnect/chip-module/CMakeLists.txt
```diff
@@ -99,9 +99,11 @@ if (CONFIG_POSIX_API)
...
+    zephyr_include_directories($<TARGET_PROPERTY:mbedcrypto_includes,INTERFACE_INCLUDE_DIRECTORIES>)
+    if(TARGET platform_cc3xx)
+    zephyr_include_directories($<TARGET_PROPERTY:platform_cc3xx,INTERFACE_INCLUDE_DIRECTORIES>)
```
tejlmand, line 105:
```suggestion
      zephyr_include_directories($<TARGET_PROPERTY:platform_cc3xx,INTERFACE_INCLUDE_DIRECTORIES>)
```

tejlmand (approved):

Damian-Nordic:
@tejlmand Good catch. So the `else` block will go away once we update the supported NCS revision in CHIP?

Damian-Nordic (approved):

tejlmand:
> @tejlmand Good catch. So the `else` block will go away once we update the supported NCS revision in CHIP?

Correct. and even the if can then go away, so the code inside the `if` is always the active code. (we will look into improving this even more in future)

I tried to update https://github.com/nrfconnect/sdk-connectedhomeip/blob/master/config/nrfconnect/.nrfconnect-recommended-revision to latest sdk-nrf/master, but seems you haven't adopted to the changes in Zephyr which came in with the latest upmerge.

This also means that `examples/*/nrfconnect` are broken in this repo when running with latest `sdk-nrf/master`, but that is outside the scope of this PR to fixup, so this is just a fyi.

Damian-Nordic:
We're working on updating the NCS revision at the moment. We were waiting for updating OpenThread, but looks like the change has finally been merged :).

PR #14, opened 2021-06-17
[nrf toup] Close BLE connection in Android CHIPTool

Damian-Nordic:
After commissioning is done, the Bluetooth LE connection
can be closed, so that subsequent devices can be paired
using the Android CHIPTool.

Signed-off-by: Damian Krolik <damian.krolik@nordicsemi.no>


LuDuda (approved):

LuDuda:
@doublemis1 is that fixes the issue?

PR #18, opened 2021-08-03
[nrf noup] add workflow for building CHIP tools

Damian-Nordic:
Add a manually triggered workflow for building Android
CHIPTool packages for arm and arm64 architectures, and
Python CHIP Controller for Linux x64 and arm64.
The packages are uploaded both as github artifacts and
release packages (as the latter require a release tag).

### DIFF
--- a/.github/workflows/release_tools.yaml
+++ b/.github/workflows/release_tools.yaml
@@ -1,0 +1,113 @@
+# Copyright (c) 2021 Project CHIP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


Damian-Nordic:
Example packages are available here: https://github.com/Damian-Nordic/sdk-connectedhomeip/releases/tag/test-release

.github/workflows/release_tools.yaml
```diff
@@ -0,0 +1,113 @@
...
+            - name: Build arm CHIPTool
+              timeout-minutes: 10
+              env:
+                TARGET_CPU: arm
+              run: |
+                  scripts/examples/android_app.sh
+                  yes | "$ANDROID_HOME/tools/bin/sdkmanager" --licenses
+                  cd src/android/CHIPTool
+                  ./gradlew build
+                  cp app/build/outputs/apk/debug/app-debug.apk /tmp/output_binaries/chiptool-arm.apk
```
LuDuda, line 89: What about to figure out a little bit more meaningful name e.g.: `chip-tool-android-arm` or sth similar?

LuDuda (commented):
This looks great, thank you! One comment about naming below.

.github/workflows/release_tools.yaml
Damian-Nordic: Fine with me. I thought `android` was implied from `apk`, but I can make it more explicit.

LuDuda (approved):

kkasperczyk-no (approved):

PR #24, opened 2021-08-24
Cherry-picked two bug fixing commits

kkasperczyk-no:
Cherry-picked two commits from upstream that contain fixes for the following bugs:
* Incorrect BLE disconnect handling and fixes for advertising switching between CHIPoBLE and SMP.
* Incorrect NFC URI scheme in Android CHIPTool app (change from CHIP to Matter CH->MT).


kkasperczyk-no:
Do not merge for now, need to cherry-pick one commit more.

PR #38, opened 2021-11-24
[nrf fromlist] Added argument allowing to bypass caching DNS-SD.

kkasperczyk-no:
DNS-SD records are cached, so in case of device IP address change
and trying to resolve address again the cached value is used
instead of the new one.

Added argument that allows to bypass reading DNS-SD services data
from cache.

### DIFF
--- a/src/lib/dnssd/Discovery_ImplPlatform.cpp
+++ b/src/lib/dnssd/Discovery_ImplPlatform.cpp
@@ -454,17 +454,20 @@ CHIP_ERROR DiscoveryImplPlatform::FinalizeServiceUpdate()
     return ChipDnssdFinalizeServiceUpdate();
 }
 
-CHIP_ERROR DiscoveryImplPlatform::ResolveNodeId(const PeerId & peerId, Inet::IPAddressType type)
+CHIP_ERROR DiscoveryImplPlatform::ResolveNodeId(const PeerId & peerId, Inet::IPAddressType type, bool dnssdCacheBypass)
 {
     ReturnErrorOnFailure(InitImpl());
 
 #if CHIP_CONFIG_MDNS_CACHE_SIZE > 0
-    /* see if the entry is cached and use it.... */
-    ResolvedNodeData nodeData;
-    if (sDnssdCache.Lookup(peerId, nodeData) == CHIP_NO_ERROR)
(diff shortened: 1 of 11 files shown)


src/lib/shell/commands/Dns.cpp
```diff
@@ -124,7 +124,7 @@ CHIP_ERROR ResolveHandler(int argc, char ** argv)
     peerId.SetCompressedFabricId(strtoull(argv[0], NULL, 10));
     peerId.SetNodeId(strtoull(argv[1], NULL, 10));
 
-    return Dnssd::Resolver::Instance().ResolveNodeId(peerId, Inet::IPAddressType::kAny);
+    return Dnssd::Resolver::Instance().ResolveNodeId(peerId, Inet::IPAddressType::kAny, true);
```
Damian-Nordic, line 127: In the upstream PR, please add `/* bypass cache */` comment next to the boolean literal. Alternatively, the community likes using enum classes for such flags, so that the code is self-explanatory. For example,
```
enum class CacheUsage
{
  On,
  Off
};
```
and then:
`Dnssd::Resolver::Instance().ResolveNodeId(peerId, Inet::IPAddressType::kAny, Dnssd::CacheUsage::On)`.

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -454,17 +454,20 @@ CHIP_ERROR DiscoveryImplPlatform::FinalizeServiceUpdate()
     return ChipDnssdFinalizeServiceUpdate();
 }
 
-CHIP_ERROR DiscoveryImplPlatform::ResolveNodeId(const PeerId & peerId, Inet::IPAddressType type)
+CHIP_ERROR DiscoveryImplPlatform::ResolveNodeId(const PeerId & peerId, Inet::IPAddressType type, bool dnssdCacheBypass)
```
Damian-Nordic, line 457: Minimal mDNS resolver must also be updated.

kkasperczyk-no:
@Damian-Nordic sdk-nrf CI is green, can we merge that one?

LuDuda (approved):

Damian-Nordic (approved):

PR #46, opened 2022-02-11
Android CHIPTool fixes

Damian-Nordic:
* Use specified PIN code when opening the commissioning window
* Fix pressure graph rendering
* Extend workflow timeout


mbolivar-nordic:
Please rebase onto the new history in master

PR #52, opened 2022-04-04
license: Replace old license reference

LuDuda:
Replace old LicenseRef with the new one currently in use.

Ref: NCSIDB-717

Signed-off-by: Lukasz Duda <lukasz.duda@nordicsemi.no>

### DIFF
--- a/zephyr/module.yml
+++ b/zephyr/module.yml
@@ -1,7 +1,7 @@
 #
 # Copyright (c) 2021 Nordic Semiconductor ASA
 #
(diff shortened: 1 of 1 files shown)


Damian-Nordic (approved):

kkasperczyk-no:
@Damian-Nordic @LuDuda can we merged that one? I would like to cherry-pick two configuration commits from the upstream, so I can update west-yml in sdk-nrf and include also this commit. Btw. commit title should be marked as [no up].

kkasperczyk-no (approved):

PR #53, opened 2022-04-06
Configuration refactor upmerge

kkasperczyk-no:


kkasperczyk-no:
For now PR is pushed for test purpose only.

PR #55, opened 2022-04-27
Upmerge 27 april

kkasperczyk-no:
Regular upmerge of Matter repository


Damian-Nordic (approved):
Nits:
- I think "Fix build with TCP..." patch may no longer be needed as it was used for the test pairing
- You can probably also squash all patches related to release_tools.yaml on the occasion to help with future upmerges. You don't have to do it now, but I think in general it's worth squashing simple fixes to simplify the downstream patch history.

PR #65, opened 2022-05-17
Added bug fixes to the Matter core

kkasperczyk-no:
Added bug fixes to the Matter core:
* QSPI sleep fix
* QR code generation fix

### DIFF
--- a/src/platform/nrfconnect/OTAImageProcessorImpl.cpp
+++ b/src/platform/nrfconnect/OTAImageProcessorImpl.cpp
@@ -36,6 +36,8 @@ CHIP_ERROR OTAImageProcessorImpl::PrepareDownload()
 {
     VerifyOrReturnError(mDownloader != nullptr, CHIP_ERROR_INCORRECT_STATE);
 
+    TriggerFlashAction(FlashHandler::Action::WAKE_UP);
+
     return DeviceLayer::SystemLayer().ScheduleLambda([this] { mDownloader->OnPreparedForDownload(PrepareDownloadImpl()); });
 }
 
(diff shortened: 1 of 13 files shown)


markaj-nordic (approved):
I think now we keep the flexibility and the FlashHandler might be overloaded by other user to implement custom flash sleep/wake up actions. Thanks @kkasperczyk-no

src/platform/nrfconnect/OTAImageProcessorImpl.cpp
```diff
@@ -206,8 +222,16 @@ CHIP_ERROR OTAImageProcessorImpl::ProcessHeader(ByteSpan & aBlock)
     return CHIP_NO_ERROR;
 }
 
+void OTAImageProcessorImpl::PutFlashSleep()
```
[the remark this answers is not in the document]
Damian-Nordic: Do we need this method at all?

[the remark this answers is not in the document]
kkasperczyk-no: So I believe it could be even SetFlashState(FlashHandler::Action state), that would be called with sleep or wake up value, right?

src/platform/nrfconnect/OTAImageProcessorImpl.cpp
```diff
@@ -67,6 +72,8 @@ CHIP_ERROR OTAImageProcessorImpl::Apply()
         err = dfu_target_schedule_update(-1);
     }
 
+    TriggerFlashAction(FlashHandler::Action::SLEEP);
```
Damian-Nordic, line 75: It's worth testing that change. I think Zigbee was fixing it, because if an app disable QSPI before rebooting, MCUboot may not be able to access it (unless our configuration fixes it in a different way).

Damian-Nordic (approved):

src/platform/nrfconnect/OTAImageProcessorImpl.cpp
kkasperczyk-no: I did single DFU and application successfully booted after swap, but we can also run nightly on it to see if we are ok.

Damian-Nordic: OK, if you tested the image is swapped then we're good. Maybe our MCUboot configuration handles that somehow :)

PR #92, opened 2022-08-11
Initial Matter over WiFi implementation

markaj-nordic:
#### Problem
This PR adds initial support for Matte-over-WiFi feature.

#### Change overview
Implemented all methods needed to commission the device to Matter running over WiFi network.
Some of the additional WiFi specific features are still under development and might not work as expected yet.
This will be fixed when the Zephyr NRF WiFi module is improved.

#### Testing
* Built the `modules/lib/matter/examples/all-clusters-app/nrfconnect` with WiFi support
* Commissioned the node to Matter network with chip-tool
* Read the security type from WiFi diagnostic cluster with `$chip-tool wifinetworkdiagnostics read security-type 1 0`


examples/all-clusters-app/nrfconnect/main/AppTask.cpp
```diff
@@ -36,6 +36,11 @@
 #include <credentials/DeviceAttestationCredsProvider.h>
 #include <credentials/examples/DeviceAttestationCredsExample.h>
 
+#ifdef CONFIG_WIFI_NRF700X
```
Damian-Nordic, line 39: Are those changes needed given that we're going to use NCS samples, anyway? If we prefer to keep it, maybe we should stick to the same samples (lock-app and light-switch) as in NCS?

Damian-Nordic (commented):
LGTM, but I think it is worth cherry-picking the test mode removal and simplify example changes.

examples/all-clusters-app/nrfconnect/main/AppTask.cpp
markaj-nordic: Hmm yes, that's doubtful. AFAIK we aim to support all-clusters-app in the upstream in the future. But you are right, we may keep this repo ncs inclined for now. Any other suggestions, @kkasperczyk-no, @LuDuda?

kkasperczyk-no: Agree, I would limit to the changes necessary for NCS. The changes to upstream examples should be somehow synced with the changes to samples in NCS, so let's wait for merging them to NCS first.

markaj-nordic: Ok, done.

Damian-Nordic (approved):

PR #97, opened 2022-08-18
[nrf noup] Fix WiFi Manager initialization

Damian-Nordic:
The WiFi Manager initialization was prone to failures due to
timing dependencies with the supplicant thread. Harden the
solution by waiting for the suppicant initialization.

Also, propagate the WiFi Manager initialization error up the
stack.


kkasperczyk-no (approved):

markaj-nordic (approved):
To me it's a workaround but we cannot do any better with existing supplicant API. Let's forward it to the WiFi team. Maybe it can be addressed in the new API design.

PR #103, opened 2022-08-29
[nrf fromtree] Update light switch doc.

markaj-nordic:
Downstreaming changes related to light switch multicast communication.
This allows usage of the simplified chip-tool command when setting up
a group communication. Also corresponding docs update.

### DIFF
--- a/examples/light-switch-app/nrfconnect/README.md
+++ b/examples/light-switch-app/nrfconnect/README.md
@@ -695,35 +695,20 @@ To perform the unicast binding process, complete the following steps:
 1.  Build the CHIP Tool according to the steps from the
     [CHIP Tool user guide](../../../docs/guides/chip_tool_guide.md#building).
 2.  Go to the CHIP Tool build directory.
-3.  Add an ACL to the [lighting endpoint](../../lighting-app/nrfconnect/)
-    permissions by running the following command:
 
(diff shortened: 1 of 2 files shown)


kkasperczyk-no (approved):
LGTM, but could you do the "git cherry-pick -x <sha>" for this upstream commit? This way it will add original commit SHA to your new commit message, so in the future it will be much easier to find it in the upstream project tree.

markaj-nordic:
> LGTM, but could you do the `git cherry-pick -x <sha> ` for this upstream commit? This way it will add original commit SHA to your new commit message, so in the future it will be much easier to find it in the upstream project tree.

The thing is, there is a difference between upstream and downstream chip-tool zap generated files. So the `cherry-pick` creates many conflicts. Therefore I needed to regenerate these files with updated .yaml configuration. And `cherry-pick` docs indicates the following "_Append the note only for cherry picks without conflicts"_ (https://linux.die.net/man/1/git-cherry-pick). So not sure if it applies here.

kkasperczyk-no:
Ok, it's up to your decision, but I don't know if I would care about it that much. Ideally `nrf fromtree` should also be 1 to 1 the commit from upstream, but it is just information for us that this commit is already in upstream tree, so we can just revert it during the next upmerge. Including original commit SHA just sometimes helps in case of problems with resolving conflicts to find it in the upstream faster.

markaj-nordic:
> Ok, it's up to your decision, but I don't know if I would care about it that much. Ideally `nrf fromtree` should also be 1 to 1 the commit from upstream, but it is just information for us that this commit is already in upstream tree, so we can just revert it during the next upmerge. Including original commit SHA just sometimes helps in case of problems with resolving conflicts to find it in the upstream faster.

@kkasperczyk-no Done.

PR #104, opened 2022-08-29
Fix build latest wifi

krish2718:
Fix the build breakage with latest Wi-Fi driver PR: https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-nrf/job/PR-8495/7/testReport/

PR #105, opened 2022-08-29
Fix build wifi offapi

krish2718:
Based on #104 but using native wifi_mgmt for doing scan rather than nrx700x specific scan. Without this once https://github.com/nrfconnect/sdk-nrf/pull/8562 this is merged, then the build will broken.


markaj-nordic:
Can be rebased with master now.

kkasperczyk-no (approved):

Damian-Nordic (approved):

PR #108, opened 2022-08-30
[nrf noup] Fixed the failsafe when connecting to WiFi wih invalid credentials.

markaj-nordic:
#### Problem
The failsafe mechanism does not work correctly when attempting to connect
to WiFi with invalid network credentials and then trying to connect with valid ones.

#### Change overview
* Reset the number of retires used when waiting for the connection.
* Remove redundant wpa_supplicant disconnect call
* Also some dead code cleanup.

#### Testing
Tested with `chip-tool`:
* `./chip-tool pairing ble-wifi 1 <invalid_ssid> <valid_password> 20202021 3840 --ble-adapter 0`
* wait until the failsafe is disarmed
* `./chip-tool pairing ble-wifi 1 <valid_ssid> <valid_password> 20202021 3840 --ble-adapter 0`
* result: the device was successfully commissioned to Matter via Wi-Fi

### DIFF
--- a/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
+++ b/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
@@ -104,13 +104,6 @@ CHIP_ERROR NrfWiFiDriver::CommitConfiguration()
 
 CHIP_ERROR NrfWiFiDriver::RevertConfiguration()
 {
-    // Disconnection should happen automatically when WiFiManager::Connect() is called.
-    // Here we do it also explicitly to ping the Connectivity Manager which
-    // will send the disconnection event as a result.
-    // TODO: refactor when the callback-based supplicant API is incorporated
-    WiFiManager::Instance().DisconnectStation();
-    ConnectivityMgr().SetWiFiStationMode(ConnectivityManager::kWiFiStationMode_Disabled);
-
     LoadFromStorage();
 
     if (mStagingNetwork.IsConfigured())
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -39,9 +39,9 @@ extern "C" {
 #include <zephyr/net/wifi_mgmt.h>
 }
 
-extern struct wpa_global *global;
+extern struct wpa_global * global;
 
-static struct wpa_supplicant *wpa_s;
+static struct wpa_supplicant * wpa_s;
 
 namespace chip {
 namespace DeviceLayer {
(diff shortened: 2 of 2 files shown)


src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -104,13 +104,6 @@ CHIP_ERROR NrfWiFiDriver::CommitConfiguration()
 
 CHIP_ERROR NrfWiFiDriver::RevertConfiguration()
 {
-    // Disconnection should happen automatically when WiFiManager::Connect() is called.
-    // Here we do it also explicitly to ping the Connectivity Manager which
-    // will send the disconnection event as a result.
-    // TODO: refactor when the callback-based supplicant API is incorporated
-    WiFiManager::Instance().DisconnectStation();
```
Damian-Nordic, line 111: Will the disconnection event be generated after removing this?

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -104,13 +104,6 @@ CHIP_ERROR NrfWiFiDriver::CommitConfiguration()
...
-    ConnectivityMgr().SetWiFiStationMode(ConnectivityManager::kWiFiStationMode_Disabled);
-
     LoadFromStorage();
 
     if (mStagingNetwork.IsConfigured())
```
Damian-Nordic, line 109: Should we disconnect if not configured?

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
markaj-nordic: It will not (actually the reason is removing the line below). But I deleted these 2 lines on purpose as DisconnectStation() call makes the wpa_supplicant thread throw a bus fault in this context (WiFi team will look at that). I mean... we can still trigger the disconnect event without direct call to disconnect but in this case we will not be sure weather or not the station really disconnects (there would have to be a corresponding event captured from supplicant first).

Also I am not sure what is the actual use case of disconnect event, we do not capture it anywhere. Just in case we really need it, it can be added once again in clean manner when the new API is delivered.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
markaj-nordic: I think we have 2 scenarios here:
- not configured and not connected (because the previous connect failed, e.g. due to invalid credentials)
- not configured, connected and the DNS-SD failed for whatever reason

For the first case we are getting mentioned bus fail, I guess it is just redundant.
I am not sure how to filter the second scenario though - in this case we should disconnect for sure.

markaj-nordic: As discussed offline, JIRA ticket (KRKNWK-14975) created for a scenario when the connection is established but the service discovery fails.

Damian-Nordic (approved):

PR #110, opened 2022-09-01
[nrf noup] Disable Wi-Fi low-power mode

Damian-Nordic:
Noticed Wi-Fi low-power mode causes performance issues,
at least with some APs.


kkasperczyk-no (approved):
Just to my curiosity, is this the reason of connection failures we observed?

greg-fer (approved):

LuDuda (approved):

Damian-Nordic:
> Just out of my curiosity, is this the reason of connection failures we observed?

It's the reason of lots of lost packets after connecting to the network.

PR #115, opened 2022-09-08
Wi-Fi network iterator fix

markaj-nordic:
#### Problem
The dummy WiFiNetworkIterator causes the infinite loop when reading
network attribute from networkcommissioning cluster.

#### Change overview
Implementation of WiFiNetworkIterator.
This PR also pulls in the BDX transfer fix from the upstream.

#### Testing
* commissioned with `chip-tool` to Matter with WiFi backend
* run `./chip-tool networkcommissioning read networks 1 0`
* received a list with one connected network

### DIFF
--- a/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
+++ b/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
@@ -32,6 +32,41 @@ namespace chip {
 namespace DeviceLayer {
 namespace NetworkCommissioning {
 
+size_t NrfWiFiDriver::WiFiNetworkIterator::Count()
+{
+    VerifyOrReturnValue(mDriver != nullptr, 0);
+    return mDriver->mStagingNetwork.IsConfigured() ? 1 : 0;
+}
+
+bool NrfWiFiDriver::WiFiNetworkIterator::Next(Network & item)
+{
+    // we assume only one network is actually supported
+    // TODO: verify if this can be extended
+    if (mExhausted || 0 == Count())
+    {
+        return false;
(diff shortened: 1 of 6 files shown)


src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -32,6 +32,41 @@ namespace chip {
...
+    item.connected    = false;
+
+    mExhausted = true;
+
+    WiFiManager::WiFiInfo wifiInfo;
+    if (CHIP_NO_ERROR == WiFiManager::Instance().GetWiFiInfo(wifiInfo))
+    {
+        if (WiFiManager::StationStatus::CONNECTED <= WiFiManager::Instance().GetStationStatus())
+        {
+            if (0 == memcmp(wifiInfo.mSsid, item.networkID, wifiInfo.mSsidLen) && wifiInfo.mSsidLen == item.networkIDLen)
```
Damian-Nordic, line 61: nit: Comparing lengths before contents would look more correct (I understand we assume that the buffer sizes are equal, but it may not be obvious for a random reader).

Damian-Nordic (approved):

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
markaj-nordic: Right, fully agree. Done.

markaj-nordic:
NCS CI is green, we can merge this PR.

kkasperczyk-no (approved):

PR #121, opened 2022-09-28
[nrf noup] Start using WiFi events instead of using wpa_supplicant

ArekBalysNordic:
#### Problem
Switching from wpa_supplicant API to WiFi mgmt events

#### Change overview
- Removed wpa_supplicant calls from WiFiManager and replaced them with WiFi events from wifi_mgmt.
- Removed a workaround from ConnectivityManagerImplWifi in _SetWiFiStationMode.
- Added ConnectionWorkerHandler to switch context during calling OnResult wifi connection callback.
- Removed a workaround from ConnectivityManagerImplWifi in _SetWiFiStationMode.

### DIFF
--- a/src/platform/nrfconnect/wifi/ConnectivityManagerImplWiFi.cpp
+++ b/src/platform/nrfconnect/wifi/ConnectivityManagerImplWiFi.cpp
@@ -57,27 +57,9 @@ ConnectivityManager::WiFiStationMode ConnectivityManagerImplWiFi::_GetWiFiStatio
 
 CHIP_ERROR ConnectivityManagerImplWiFi::_SetWiFiStationMode(ConnectivityManager::WiFiStationMode aMode)
 {
-    VerifyOrReturnError(ConnectivityManager::WiFiStationMode::kWiFiStationMode_NotSupported != aMode,
-                        CHIP_ERROR_UNSUPPORTED_CHIP_FEATURE);
+    VerifyOrReturnError(ConnectivityManager::WiFiStationMode::kWiFiStationMode_NotSupported != aMode, CHIP_ERROR_INVALID_ARGUMENT);
 
-    if (aMode != mStationMode)
-    {
-        mStationMode = aMode;
-        if (mStationMode != ConnectivityManager::WiFiStationMode::kWiFiStationMode_ApplicationControlled)
-        {
-            bool doEnable{ ConnectivityManager::WiFiStationMode::kWiFiStationMode_Enabled == mStationMode };
-            // TODO: when the connection/disconnection callback API is provided
-            // below calls should be used as a base of disconnect callback
-            if (doEnable)
-            {
-                OnStationConnected();
-            }
-            else
-            {
-                OnStationDisconnected();
-            }
-        }
-    }
+    mStationMode = aMode;
 
     return CHIP_NO_ERROR;
 }
@@ -94,7 +76,7 @@ bool ConnectivityManagerImplWiFi::_IsWiFiStationApplicationControlled(void)
 
 bool ConnectivityManagerImplWiFi::_IsWiFiStationConnected(void)
 {
-    return (WiFiManager::StationStatus::FULLY_PROVISIONED == WiFiManager().Instance().GetStationStatus());
+    return (WiFiManager::StationStatus::CONNECTED == WiFiManager().Instance().GetStationStatus());
 }
 
 System::Clock::Timeout ConnectivityManagerImplWiFi::_GetWiFiStationReconnectInterval(void)
@@ -144,38 +126,6 @@ CHIP_ERROR ConnectivityManagerImplWiFi::_GetAndLogWiFiStatsCounters(void)
 {
     // TODO: when network statistics are enabled
     return CHIP_NO_ERROR;
-}
-
-void ConnectivityManagerImplWiFi::OnStationConnected()
-{
-    // ensure the station is connected
-    if (_IsWiFiStationConnected())
-    {
-        ChipDeviceEvent connectEvent{};
-        connectEvent.Type                          = DeviceEventType::kWiFiConnectivityChange;
-        connectEvent.WiFiConnectivityChange.Result = kConnectivity_Established;
-        PlatformMgr().PostEventOrDie(&connectEvent);
-    }
-    else
-    {
-        ChipLogError(DeviceLayer, "WiFi Station is not connected!");
-    }
-}
-
-void ConnectivityManagerImplWiFi::OnStationDisconnected()
-{
-    // ensure the station is disconnected
-    if (WiFiManager::StationStatus::DISCONNECTED == WiFiManager().Instance().GetStationStatus())
-    {
-        ChipDeviceEvent disconnectEvent{};
-        disconnectEvent.Type                          = DeviceEventType::kWiFiConnectivityChange;
-        disconnectEvent.WiFiConnectivityChange.Result = kConnectivity_Lost;
-        PlatformMgr().PostEventOrDie(&disconnectEvent);
-    }
-    else
-    {
-        ChipLogError(DeviceLayer, "WiFi Station is not disconnected!");
-    }
 }
 
 ConnectivityManager::WiFiAPMode ConnectivityManagerImplWiFi::_GetWiFiAPMode(void)
--- a/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
+++ b/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
@@ -17,7 +17,6 @@
 
 #include "NrfWiFiDriver.h"
 
-#include "WiFiManager.h"
 #include <platform/KeyValueStoreManager.h>
 
 #include <lib/support/CodeUtils.h>
@@ -27,6 +26,7 @@
 using namespace ::chip;
 using namespace ::chip::DeviceLayer::Internal;
 using namespace ::chip::DeviceLayer::PersistedStorage;
+using namespace ::chip::app::Clusters::NetworkCommissioning;
 
 namespace chip {
 namespace DeviceLayer {
@@ -105,7 +105,7 @@ CHIP_ERROR NrfWiFiDriver::Init(NetworkStatusChangeCallback * networkStatusChange
     {
         WiFiManager::ConnectionHandling handling{ [] { Instance().OnNetworkStatusChanged(Status::kSuccess); },
                                                   [] { Instance().OnNetworkStatusChanged(Status::kUnknownError); },
-                                                  System::Clock::Timeout{ 40000 } };
+                                                  System::Clock::Timeout{ kWiFiConnectNetworkTimeoutSeconds * 1000 } };
         ReturnErrorOnFailure(
             WiFiManager::Instance().Connect(mStagingNetwork.GetSsidSpan(), mStagingNetwork.GetPassSpan(), handling));
     }
@@ -121,7 +121,15 @@ void NrfWiFiDriver::OnNetworkStatusChanged(Status status)
     }
 
     if (mpNetworkStatusChangeCallback)
+    {
         mpNetworkStatusChangeCallback->OnNetworkingStatusChange(status, NullOptional, NullOptional);
+    }
+
+    if (mpConnectCallback)
+    {
+        mpConnectCallback->OnResult(status, CharSpan(), 0);
+        mpConnectCallback = nullptr;
+    }
 }
 
 void NrfWiFiDriver::Shutdown()
@@ -143,9 +151,9 @@ CHIP_ERROR NrfWiFiDriver::RevertConfiguration()
 
     if (mStagingNetwork.IsConfigured())
     {
-        WiFiManager::ConnectionHandling handling{ [] { Instance().OnConnectWiFiNetwork(); },
-                                                  [] { Instance().OnConnectWiFiNetworkFailed(); },
-                                                  System::Clock::Timeout{ 40000 } };
+        WiFiManager::ConnectionHandling handling{ [] { Instance().OnNetworkStatusChanged(Status::kSuccess); },
+                                                  [] { Instance().OnNetworkStatusChanged(Status::kUnknownError); },
+                                                  System::Clock::Timeout{ kWiFiConnectNetworkTimeoutSeconds * 1000 } };
         ReturnErrorOnFailure(
             WiFiManager::Instance().Connect(mStagingNetwork.GetSsidSpan(), mStagingNetwork.GetPassSpan(), handling));
     }
@@ -196,9 +204,12 @@ Status NrfWiFiDriver::ReorderNetwork(ByteSpan networkId, uint8_t index, MutableC
 void NrfWiFiDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * callback)
 {
     Status status = Status::kSuccess;
-    WiFiManager::ConnectionHandling handling{ [] { Instance().OnConnectWiFiNetwork(); },
-                                              [] { Instance().OnConnectWiFiNetworkFailed(); }, System::Clock::Timeout{ 40000 } };
+    WiFiManager::ConnectionHandling handling{ [] { Instance().OnNetworkStatusChanged(Status::kSuccess); },
+                                              [] { Instance().OnNetworkStatusChanged(Status::kUnknownError); },
+                                              System::Clock::Timeout{ kWiFiConnectNetworkTimeoutSeconds * 1000 } };
 
+    VerifyOrExit(WiFiManager::StationStatus::CONNECTING != WiFiManager::Instance().GetStationStatus(),
+                 status = Status::kOtherConnectionFailure);
     VerifyOrExit(networkId.data_equal(mStagingNetwork.GetSsidSpan()), status = Status::kNetworkIDNotFound);
     VerifyOrExit(mpConnectCallback == nullptr, status = Status::kUnknownError);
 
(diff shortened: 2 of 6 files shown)


src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -27,6 +28,9 @@
 using namespace ::chip;
 using namespace ::chip::DeviceLayer::Internal;
 using namespace ::chip::DeviceLayer::PersistedStorage;
+using namespace ::chip::app::Clusters::NetworkCommissioning;
+
+constexpr static auto WiFiCommissioningTimeoutMs = 10000;
```
kkasperczyk-no, line 33: 1. WiFiCommissioningTimeoutMs-> kWiFiCommissioningTimeoutMs
2. Can we move it to the NrfWifiDriver.h, as a class field?

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -209,74 +220,103 @@ void NrfWiFiDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * callbac
     if (status != Status::kSuccess)
     {
         mpConnectCallback = nullptr;
-        callback->OnResult(status, CharSpan(), 0);
+
+        NetworkCommissioningStatus * data = Platform::New<NetworkCommissioningStatus>(status);
```
kkasperczyk-no, line 224: Shouldn't we check if data was successfully allocated before using it?

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -209,74 +220,103 @@ void NrfWiFiDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * callbac
...
+        mScanCallback->OnFinished(status == 0 ? Status::kSuccess : Status::kUnknownError, CharSpan(), &mScanResponseIterator);
+        mScanCallback = nullptr;
+    });
 }
 
-void NrfWiFiDriver::OnConnectWiFiNetworkFailed()
+void NrfWiFiDriver::OnConnectNetwork(Status status)
 {
-    if (mpConnectCallback)
+    NetworkCommissioningStatus * data = Platform::New<NetworkCommissioningStatus>(status);
```
kkasperczyk-no, line 258: Ditto

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -209,74 +220,103 @@ void NrfWiFiDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * callbac
...
+        networkEvent.WiFiConnectivityChange.Result = kConnectivity_Established;
+    }
+    else if (WiFiManager::StationStatus::DISCONNECTED == WiFiManager().Instance().GetStationStatus())
+    {
+        networkEvent.Type                          = DeviceEventType::kWiFiConnectivityChange;
+        networkEvent.WiFiConnectivityChange.Result = kConnectivity_Lost;
+    }
+    else
+    {
+        ChipLogError(DeviceLayer, "WiFi is not responding!");
```
kkasperczyk-no, line 302: Can we add to this message what is the current status code?

src/platform/nrfconnect/wifi/WiFiManager.h
```diff
@@ -140,31 +145,31 @@ class WiFiManager
...
+    static void mConnectClbk(struct net_mgmt_event_callback * cb);
+    static void mDisconnectClbk(struct net_mgmt_event_callback * cb);
 
     static uint8_t FrequencyToChannel(uint16_t freq);
-    static StationStatus StatusFromWpaStatus(const wpa_states & status);
 
-    static const Map<wpa_states, StationStatus, 10> sStatusMap;
-    static const Map<uint16_t, uint8_t, 42> sFreqChannelMap;
+    struct wifi_connect_req_params mWiFiParams = {};
+    static uint32_t mScanResult;
```
kkasperczyk-no, line 168: I think it shouldn't be static. If you want to access it from the static methods, you can create static method to get WiFiManager object instance and access this field that way.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
ArekBalysNordic: Hmm or maybe we should read it from a specific cluster? What do you think, @markaj-nordic ?

src/platform/nrfconnect/wifi/WiFiManager.h
markaj-nordic: The `WiFiManager` is a singleton anyway. So Instance().mScanResult should do the job with mScanResult being non-static class member. BTW. if static, the `m` prefix does not apply here.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
markaj-nordic: The NrfWiFiDriver.h already has a `constexpr uint8_t kWiFiConnectNetworkTimeoutSeconds = 120` which is used to set the corresponding cluster read by the controller during the commissioning. I guess we can end up with only one const for both cases and convert seconds/ms if needed. Also note that this is a **connection** timeout not **commissioning** timeout. So the naming should be `kWiFiConnectionTimeoutMs`

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -145,302 +139,203 @@ CHIP_ERROR WiFiManager::Init()
...
-CHIP_ERROR WiFiManager::AddNetwork(const ByteSpan & ssid, const ByteSpan & credentials)
+CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanCallback callback)
 {
-    ChipLogDetail(DeviceLayer, "Adding WiFi network");
-    mpWpaNetwork = wpa_supplicant_add_network(wpa_s);
-    if (mpWpaNetwork)
-    {
-        static constexpr size_t kMaxSsidLen{ 32 };
-        mpWpaNetwork->ssid = (u8 *) k_malloc(kMaxSsidLen);
+    struct net_if * iface = net_if_get_default();
```
markaj-nordic, line 151: Please use the GetInterface() helper function from Zephyr utils whenever accessing the net iface:
https://github.com/nrfconnect/sdk-connectedhomeip/blob/238a18db4eae15381c3c1838d8e863d841c59ead/src/platform/Zephyr/InetUtils.cpp#L32

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -73,49 +73,43 @@ NetworkCommissioning::WiFiScanResponse ToScanResponse(wifi_scan_result * result)
...
+                              { WIFI_STATE_INACTIVE, WiFiManager::StationStatus::DISABLED },
+                              { WIFI_STATE_SCANNING, WiFiManager::StationStatus::SCANNING },
+                              { WIFI_STATE_AUTHENTICATING, WiFiManager::StationStatus::CONNECTING },
+                              { WIFI_STATE_ASSOCIATING, WiFiManager::StationStatus::CONNECTING },
+                              { WIFI_STATE_ASSOCIATED, WiFiManager::StationStatus::CONNECTED },
+                              { WIFI_STATE_4WAY_HANDSHAKE, WiFiManager::StationStatus::PROVISIONING },
+                              { WIFI_STATE_GROUP_HANDSHAKE, WiFiManager::StationStatus::PROVISIONING },
+                              { WIFI_STATE_COMPLETED, WiFiManager::StationStatus::FULLY_PROVISIONED } });
+
+struct net_mgmt_event_callback WiFiManager::mWiFiMgmtClbk = {};
```
markaj-nordic, line 88: Do we need it here?

ArekBalysNordic: I guess so because this is a static member of WiFiManager class and we should define it.

src/platform/nrfconnect/wifi/WiFiManager.cpp
ArekBalysNordic: How can I obtain ifaceId there?

src/platform/nrfconnect/wifi/WiFiManager.h
```diff
@@ -87,6 +88,9 @@ class WiFiManager
 {
     using ConnectionCallback = void (*)();
```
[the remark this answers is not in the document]
markaj-nordic: Sorry, by bad :)

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: You don't have to, the default argument is nullptr. Eventually it boils down to calling `net_if_get_default() `anyway, but we should have the unified way of obtaining the iface.

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Sorry, I should've formulated the question more clearly. I mean it doesn't have to be a static member. And even if, we can use the Instance() method to get it, for example inside a static method.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -145,302 +139,203 @@ CHIP_ERROR WiFiManager::Init()
...
+    net_mgmt_init_event_callback(&mWiFiMgmtClbk, WifiMgmtEventHandler, sWifiManagementEvents);
+    net_mgmt_add_event_callback(&mWiFiMgmtClbk);
+
+    ChipLogDetail(DeviceLayer, "WiFiManager has been initialized");
 
     return CHIP_NO_ERROR;
 }
-
-CHIP_ERROR WiFiManager::AddNetwork(const ByteSpan & ssid, const ByteSpan & credentials)
+CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanCallback callback)
```
markaj-nordic, line 149: Looks like the ScanCallback is not registered and the scan response is not translated to Matter data types. Therefore the scan command sent from the controller will not work. I will try to fix that.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -209,74 +218,110 @@ void NrfWiFiDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * callbac
...
 
-void NrfWiFiDriver::OnScanWiFiNetworkDone(int status, WiFiScanResponse * response)
+void NrfWiFiDriver::PostConnectivityStatusChange()
 {
-    if (response != nullptr)
+    ChipDeviceEvent networkEvent{};
+
+    WiFiManager::StationStatus status = WiFiManager::Instance().GetStationStatus();
+
+    if (WiFiManager::StationStatus::DISABLED <= status)
```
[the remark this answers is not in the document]
markaj-nordic: Fixed

src/platform/nrfconnect/wifi/ConnectivityManagerImplWiFi.cpp
```diff
@@ -63,20 +63,6 @@ CHIP_ERROR ConnectivityManagerImplWiFi::_SetWiFiStationMode(ConnectivityManager:
     if (aMode != mStationMode)
```
Damian-Nordic, line 63: 1. This `if` seems unnecessary now.
2. Also, `CHIP_ERROR_INVALID_ARGUMENT` instead of `CHIP_ERROR_UNSUPPORTED_CHIP_FEATURE` should be used above.
3. Note that semantics of this method is different than on other platforms (which actually enable or disable the Wi-Fi interface). It doesn't matter a lot as long as it's not called by the application, but it can be surprising to customers. May be a good idea to file a task to sort it out and align the behavior with other platforms.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -163,6 +165,10 @@ Status NrfWiFiDriver::AddOrUpdateNetwork(ByteSpan ssid, ByteSpan credentials, Mu
     VerifyOrReturnError(ssid.size() <= sizeof(mStagingNetwork.ssid), Status::kOutOfRange);
     VerifyOrReturnError(credentials.size() <= sizeof(mStagingNetwork.pass), Status::kOutOfRange);
 
+    // It is forbidden to change network credentials during associating process.
```
Damian-Nordic, line 168: Should it rather be checked in `ConnectNetwork`? This function only updates the local credentials object. What shouldn't happen is calling WPA supplicant API while processing the previous request, right?

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -206,77 +213,94 @@ void NrfWiFiDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * callbac
     WiFiManager::Instance().Connect(mStagingNetwork.GetSsidSpan(), mStagingNetwork.GetPassSpan(), handling);
 
 exit:
-    if (status != Status::kSuccess)
+    if (Instance().mpConnectCallback)
```
Damian-Nordic, line 216: `Instance()` is unnecessary. Also, shouldn't we only call it on failure? Otherwise, `Connect` is responsible for calling the callback, no?

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -206,77 +213,94 @@ void NrfWiFiDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * callbac
...
+    WiFiNetwork network;
+
+    mStagingNetwork = {};
+    ReturnOnFailure(KeyValueStoreMgr().Get(kSsidKey, network.ssid, sizeof(network.ssid), &network.ssidLen));
+    ReturnOnFailure(KeyValueStoreMgr().Get(kPassKey, network.pass, sizeof(network.pass), &network.passLen));
+    mStagingNetwork = network;
 }
 
-void NrfWiFiDriver::OnConnectWiFiNetwork()
+void NrfWiFiDriver::OnScanWiFiNetworkDone(WiFiManager::WiFiRequestStatus status)
```
Damian-Nordic, line 233: Is it called in the context of CHIP thread? `OnScanWiFiNetworkResult` suggests it's not, in which case `ScheduleLambda` may be needed.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -206,77 +213,94 @@ void NrfWiFiDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * callbac
...
+        networkEvent.WiFiConnectivityChange.Result = kConnectivity_Established;
+    }
+    else if (WiFiManager::StationStatus::DISCONNECTED == status)
+    {
+        networkEvent.Type                          = DeviceEventType::kWiFiConnectivityChange;
+        networkEvent.WiFiConnectivityChange.Result = kConnectivity_Lost;
+    }
+    else
+    {
+        ChipLogError(DeviceLayer, "WiFi is not responding! Current Status: %d", static_cast<uint8_t>(status));
```
Damian-Nordic, line 285: The message is unclear to me. What exactly is not responding? WPA supplicant?

src/platform/nrfconnect/wifi/ConnectivityManagerImplWiFi.cpp
markaj-nordic: 1. Done.
2. Done.
3. I agree, originally I wanted to stick to the semantics similar to other platforms (although slightly forcefully). Currently it makes not much sense to do it, but yes, we should check if this can lead do some potential problems on customer side and better understand the role of ConnectivityManager entity. JIRA ticket created.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
markaj-nordic: Yes... I guess this came from wrong understanding of what this method should really do. Fixed.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
markaj-nordic: `OnScanWiFiNetworkResult` is registered (with the use of lambda) as a net_mgmt event callback in `WiFiManager::Scan()`
And with my recent commit all net_mgmt event callbacks are handled in CHIP context (`WifiMgmtEventHandler` -> `WifiMgmtEventHandlerImpl`). But please check again, I may be missing something.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
Damian-Nordic: I can't see any "return" after `Connect`, so that didn't seem to be the case.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
Damian-Nordic: Ok, then `OnScanWiFiNetworkResult` probably shouldn't use `StackLock`, right?

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
markaj-nordic: Ahh right, thanks for the catch :)

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -73,49 +73,46 @@ NetworkCommissioning::WiFiScanResponse ToScanResponse(wifi_scan_result * result)
...
+
+const Map<uint32_t, WiFiManager::NetEventHandler, 4>
+    WiFiManager::sEventHandlerMap({ { NET_EVENT_WIFI_SCAN_RESULT, WiFiManager::ScanResultClbk },
+                                    { NET_EVENT_WIFI_SCAN_DONE, WiFiManager::ScanDoneClbk },
+                                    { NET_EVENT_WIFI_CONNECT_RESULT, WiFiManager::ConnectClbk },
+                                    { NET_EVENT_WIFI_DISCONNECT_RESULT, WiFiManager::DisconnectClbk } });
+
+void WiFiManager::WifiMgmtEventHandler(NetEventCallback * cb, uint32_t mgmtEvent, struct net_if * iface)
+{
+    NetworkEventData * data = Platform::New<NetworkEventData>();
```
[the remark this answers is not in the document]
markaj-nordic: > Wouldn't SystemLayer::ScheduleLambda be more type-safe and remove the need for memory allocation?
Yep, that came into my mind but I was not sure if this can be used for the same purpose as the namespaces are different (Platform vs SystemLayer)

> Also, filter by Wi-Fi interface or add a TODO to do so.
Ok, will do.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
markaj-nordic: Yes, sorry, I have overlooked that.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -122,6 +122,8 @@ void NrfWiFiDriver::OnNetworkStatusChanged(Status status)
 
     if (mpNetworkStatusChangeCallback)
         mpNetworkStatusChangeCallback->OnNetworkingStatusChange(status, NullOptional, NullOptional);
+
```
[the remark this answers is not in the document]
markaj-nordic: There two use different callbacks under the hood, but indeed I see no reason why these cannot be fired simultaneously. I'll try to merge as suggested.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -206,77 +213,94 @@ void NrfWiFiDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * callbac
...
     }
+
+    PostConnectivityStatusChange();
 }
 
-void NrfWiFiDriver::OnScanWiFiNetworkDone(int status, WiFiScanResponse * response)
+void NrfWiFiDriver::PostConnectivityStatusChange()
 {
-    if (response != nullptr)
+    ChipDeviceEvent networkEvent{};
```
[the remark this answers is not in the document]
markaj-nordic: Hmm true. Let's try to move it to the `WiFiManager::DisconnectClbk`.

Damian-Nordic (approved):
LGTM!

PR #123, opened 2022-10-07
[nrf noup] Implemented WiFi security modes handling

markaj-nordic:
#### Problem
We need proper security modes handling to allow our devices to connect to AP supporting various security modes.

#### Change overview
Utilized net_mgmt API to perform scanning before the actual connection request.
By doing so, we can cache the security mode advertised by every AP in the area
and filter the interested one by SSID.

Also fixed the bug with incorrect net_mgmt events lifetime management.

#### Testing
Tested by commissioning the light switch app to the networks with:
* WPA2 
* open security mode
* WPA3 (tested with iPhone working as an access point)

### DIFF
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -86,16 +86,32 @@ const Map<wifi_iface_state, WiFiManager::StationStatus, 10>
                               { WIFI_STATE_COMPLETED, WiFiManager::StationStatus::FULLY_PROVISIONED } });
 
 const Map<uint32_t, WiFiManager::NetEventHandler, 4>
-    WiFiManager::sEventHandlerMap({ { NET_EVENT_WIFI_SCAN_RESULT, WiFiManager::ScanResultClbk },
-                                    { NET_EVENT_WIFI_SCAN_DONE, WiFiManager::ScanDoneClbk },
-                                    { NET_EVENT_WIFI_CONNECT_RESULT, WiFiManager::ConnectClbk },
-                                    { NET_EVENT_WIFI_DISCONNECT_RESULT, WiFiManager::DisconnectClbk } });
+    WiFiManager::sEventHandlerMap({ { NET_EVENT_WIFI_SCAN_RESULT, WiFiManager::ScanResultHandler },
+                                    { NET_EVENT_WIFI_SCAN_DONE, WiFiManager::ScanDoneHandler },
+                                    { NET_EVENT_WIFI_CONNECT_RESULT, WiFiManager::ConnectHandler },
+                                    { NET_EVENT_WIFI_DISCONNECT_RESULT, WiFiManager::DisconnectHandler } });
 
-void WiFiManager::WifiMgmtEventHandler(NetEventCallback * cb, uint32_t mgmtEvent, struct net_if * iface)
+void WiFiManager::WifiMgmtEventHandler(net_mgmt_event_callback * cb, uint32_t mgmtEvent, net_if * iface)
 {
     if (0 == strcmp(iface->if_dev->dev->name, "wlan0"))
     {
-        DeviceLayer::SystemLayer().ScheduleLambda([cb, mgmtEvent] { sEventHandlerMap[mgmtEvent](cb); });
+        Platform::UniquePtr<uint8_t> eventData(new uint8_t[cb->info_length]);
+        VerifyOrReturn(eventData);
+        memcpy(eventData.get(), cb->info, cb->info_length);
+        CHIP_ERROR status = SystemLayer().ScheduleLambda([data = eventData.get(), mgmtEvent]() {
+            if (data)
+            {
+                sEventHandlerMap[mgmtEvent](data);
+                // cleanup
+                delete[] data;
+            }
+        });
+
+        if (CHIP_NO_ERROR == status)
+        {
+            // the ownership has been transferred to the worker thread - release the buffer
+            eventData.release();
+        }
     }
 }
 
@@ -137,11 +153,13 @@ CHIP_ERROR WiFiManager::Init()
 
     return CHIP_NO_ERROR;
 }
-CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCallback, ScanDoneCallback doneCallback)
+CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCallback, ScanDoneCallback doneCallback,
+                             bool internalScan)
 {
-    struct net_if * iface = InetUtils::GetInterface();
+    net_if * iface = InetUtils::GetInterface();
     VerifyOrReturnError(nullptr != iface, CHIP_ERROR_INTERNAL);
 
+    mInternalScan       = internalScan;
     mScanResultCallback = resultCallback;
     mScanDoneCallback   = doneCallback;
     mWiFiState          = WIFI_STATE_SCANNING;
@@ -159,7 +177,8 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
 
 CHIP_ERROR WiFiManager::ClearStationProvisioningData()
 {
-    memset(&mWiFiParams, 0, sizeof(mWiFiParams));
+    mWiFiParams.mRssi = std::numeric_limits<int8_t>::min();
+    memset(&mWiFiParams.mParams, 0, sizeof(mWiFiParams.mParams));
     return CHIP_NO_ERROR;
 }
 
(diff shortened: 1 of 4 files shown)


src/platform/nrfconnect/wifi/WiFiManager.h
```diff
@@ -155,31 +156,33 @@ class WiFiManager
     CHIP_ERROR GetNetworkStatistics(NetworkStatistics & stats) const;
 
 private:
-    using ConnectionParams = wifi_connect_req_params;
     using NetEventCallback = net_mgmt_event_callback;
     using WiFiStatus       = wifi_status;
-    using NetEventHandler  = void (*)(NetEventCallback *);
+    using NetInterfacePtr  = net_if *;
```
Damian-Nordic, line 161: Maybe it's just me, but IMHO aliasing Zephyr types isn't really useful as the code isn't generic enough to allow for easy swap of the underlying type, and it makes the code harder to follow, because I need to keep the mapping in mind to verify if we set all the fields properly. It makes sense in the case of callbacks, but if the original type name is self-describing, maybe "entities should not be multiplied beyond necessity"?
But let others comment on that, too :)

markaj-nordic: Yehhh... It made more sense when I was using the `struct` here which was redundant.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -86,19 +86,36 @@ const Map<wifi_iface_state, WiFiManager::StationStatus, 10>
...
-void WiFiManager::WifiMgmtEventHandler(NetEventCallback * cb, uint32_t mgmtEvent, struct net_if * iface)
+void WiFiManager::WifiMgmtEventHandler(NetEventCallback * cb, uint32_t mgmtEvent, NetInterfacePtr iface)
 {
     if (0 == strcmp(iface->if_dev->dev->name, "wlan0"))
     {
-        DeviceLayer::SystemLayer().ScheduleLambda([cb, mgmtEvent] { sEventHandlerMap[mgmtEvent](cb); });
+        NetworkEventData * eventData = new NetworkEventData;
+        VerifyOrReturn(nullptr != eventData);
+        uint8_t * info = new uint8_t[cb->info_length];
+        VerifyOrReturn(nullptr != info);
```
Damian-Nordic, line 101: Please use `Platform::UniquePtr` to store the pointer and call `release()` on it only if the `ScheduleWork` or `ScheduleLambda` and everything else succeeds (not to leak memory on failure).

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -86,19 +86,36 @@ const Map<wifi_iface_state, WiFiManager::StationStatus, 10>
...
+                                    { NET_EVENT_WIFI_CONNECT_RESULT, WiFiManager::ConnectHandler },
+                                    { NET_EVENT_WIFI_DISCONNECT_RESULT, WiFiManager::DisconnectHandler } });
 
-void WiFiManager::WifiMgmtEventHandler(NetEventCallback * cb, uint32_t mgmtEvent, struct net_if * iface)
+void WiFiManager::WifiMgmtEventHandler(NetEventCallback * cb, uint32_t mgmtEvent, NetInterfacePtr iface)
 {
     if (0 == strcmp(iface->if_dev->dev->name, "wlan0"))
     {
-        DeviceLayer::SystemLayer().ScheduleLambda([cb, mgmtEvent] { sEventHandlerMap[mgmtEvent](cb); });
+        NetworkEventData * eventData = new NetworkEventData;
```
Damian-Nordic, line 98: This could be avoided - you could pass contents of this structure to a lambda passed to `ScheduleLambda` and get rid off one `reinterpret_cast` at least. But if you prefer it this way please use `Platform::UniquePtr` not to leak memory on failures.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -169,38 +189,22 @@ CHIP_ERROR WiFiManager::Connect(const ByteSpan & ssid, const ByteSpan & credenti
...
-        if (Instance().mHandling.mOnConnectionFailed)
-            Instance().mHandling.mOnConnectionFailed();
-        return CHIP_ERROR_INTERNAL;
-    }
-
-    ChipLogError(DeviceLayer, "Connection to %s requested", ssid.data());
-
+    // Store SSID and credentials and perform the scan to detect the security mode supported by the AP.
+    // Connect() will be called in the callback when we have the SSID match.
+    mCurrentCredentials = credentials;
```
Damian-Nordic, line 198: Who is responsible for allocating/deallocating memory referenced by these spans?

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -169,38 +189,22 @@ CHIP_ERROR WiFiManager::Connect(const ByteSpan & ssid, const ByteSpan & credenti
...
+    mWantedSSID         = ssid;
+    CHIP_ERROR err      = Scan(ssid, nullptr, nullptr, true /* internal scan */);
+    VerifyOrReturnError(CHIP_NO_ERROR == err, err);
     return CHIP_NO_ERROR;
```
Damian-Nordic, line 202:
```suggestion
    return Scan(ssid, nullptr, nullptr, true /* internal scan */);
```

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -272,35 +276,91 @@ CHIP_ERROR WiFiManager::GetNetworkStatistics(NetworkStatistics & stats) const
...
+void WiFiManager::ScanResultHandler(uint8_t * data)
 {
-    VerifyOrReturn(Instance().mScanResultCallback != nullptr);
-    const struct wifi_scan_result * scanResult      = (const struct wifi_scan_result *) cb->info;
-    NetworkCommissioning::WiFiScanResponse response = ToScanResponse(scanResult);
-    Instance().mScanResultCallback(scanResult != nullptr ? &response : nullptr);
+    const struct wifi_scan_result * scanResult = reinterpret_cast<const struct wifi_scan_result *>(data);
+
+    if (Instance().mInternalScan &&
+        strcmp(reinterpret_cast<const char *>(scanResult->ssid), reinterpret_cast<const char *>(Instance().mWantedSSID.data())) ==
```
Damian-Nordic, line 284: How do we know it's null-terminated. Shouldn't we rather use `ssid_length`? Note that `ByteSpan` has `data_equals` method to ease such comparison.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -272,35 +276,91 @@ CHIP_ERROR WiFiManager::GetNetworkStatistics(NetworkStatistics & stats) const
...
+            Instance().mWiFiParams.mParams.timeout = Instance().mHandling.mConnectionTimeoutMs.count();
+            Instance().mWiFiParams.mParams.channel = scanResult->channel;
+            Instance().mWiFiParams.mRssi           = scanResult->rssi;
+        }
+    }
+
+    if (Instance().mScanResultCallback && !Instance().mInternalScan)
+    {
+        NetworkCommissioning::WiFiScanResponse response = ToScanResponse(scanResult);
+        Instance().mScanResultCallback(scanResult != nullptr ? &response : nullptr);
```
Damian-Nordic, line 318: Is it possible that it's null? Above you seem to assume it is not. In the old API nullptr was used to indicate `done` callbacks, but maybe it's now unnecessary and `mScanResultCallback` could take a reference.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -272,35 +276,91 @@ CHIP_ERROR WiFiManager::GetNetworkStatistics(NetworkStatistics & stats) const
...
+
+            if (net_mgmt(NET_REQUEST_WIFI_CONNECT, iface, &(Instance().mWiFiParams.mParams), sizeof(wifi_connect_req_params)))
+            {
+                ChipLogError(DeviceLayer, "Connection request failed");
+                if (Instance().mHandling.mOnConnectionFailed)
+                {
+                    Instance().mHandling.mOnConnectionFailed();
+                }
+            }
+            ChipLogError(DeviceLayer, "Connection to %s requested", Instance().mWiFiParams.mParams.ssid);
```
Damian-Nordic, line 355: Is ssid null-terminated?

Damian-Nordic, line 355: This will be printed even in the case of error. Missing `return` above?

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: My first approach was to pass the ownership to the handler by moving the `std::unique_ptr` with data to the `ScheduleLambda` (in the capture scope)  - that would be a good semantics here IMO but it appears that the API expects trivially copyable lambda, so we cannot capture by move.

So how can we actually use the `Platform::UniquePtr` here? By passing underlying raw pointer directly to the lambda? Don't we need two allocations here anyway meaning that the `mData` inside `NetworkEventData` should be a unique_ptr as well?

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Using `ScheduleLambda` was my first idea, Please see the comment above.

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Good catch, thanks.

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Right, I guess reference can be used here. Will fix that.

src/platform/nrfconnect/wifi/WiFiManager.cpp
Damian-Nordic: My point is that we should use the unique pointer only to protect the buffer from leaking. Once the buffer is passed to another thread we can release the unique pointer.

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Hmm not sure, probably we cannot assume that indeed. I will use the `%*s` formatting

markaj-nordic: Yes, thanks.

src/platform/nrfconnect/wifi/WiFiManager.cpp
Damian-Nordic: We can still remove `NetworkEventData` and pass info as an allocated buffer, right?

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Hmm good question. As far as I understand the ByteSpan is just a wrapper for a raw pointer and data length and it does not do any deep copy under the hood. So essentially these spans (a.k.a underlying buffers) are owned by the `NrfWiFiDriver' instance (staging network object). Sounds like sth error prone. Maybe better have a local copy.

Damian-Nordic: Yes, I think we should copy them or document the ownership explicitly.

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Ok, done.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -86,16 +86,32 @@ const Map<wifi_iface_state, WiFiManager::StationStatus, 10>
...
+void WiFiManager::WifiMgmtEventHandler(net_mgmt_event_callback * cb, uint32_t mgmtEvent, net_if * iface)
 {
     if (0 == strcmp(iface->if_dev->dev->name, "wlan0"))
     {
-        DeviceLayer::SystemLayer().ScheduleLambda([cb, mgmtEvent] { sEventHandlerMap[mgmtEvent](cb); });
+        Platform::UniquePtr<uint8_t> eventData(new uint8_t[cb->info_length]);
+        VerifyOrReturn(eventData);
+        memcpy(eventData.get(), cb->info, cb->info_length);
+        CHIP_ERROR status = SystemLayer().ScheduleLambda([data = eventData.get(), mgmtEvent]() {
+            if (data)
```
Damian-Nordic, line 102: It must be true, right?

Damian-Nordic (approved):
Looks great, thanks!

LuDuda (approved):
Great work :+1: Thanks!

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Yes, it's redundant, but also does not harm :) This is happening in separate thread and the event handlers do not check for nullptr so maybe let's keep it for now.

PR #132, opened 2022-10-18
[nrf noup] WiFi fail-safe related fixes

markaj-nordic:
Make sure the fail-safe works as expected with WiFi networking:
- disconnect when reverting the configuration
- always cleanup the provisioning data structures before connecting to avoid dummy buffer overwriting and data length issues

Tested the following scenarios manually with `chip-tool`:

**Commissioning (failure)**
1. GeneralCommissioning.ArmFailSafe
2. NetworkCommissioning.AddOrUpdateWifiNetwork
3. NetworkCommissioning.ConnectNetwork
4. DNS-SD -> failure
5. Failsafe timeout - RevertConfiguration - disconnect from tentative Wi-Fi network

**Commissioning (success)**
1. GeneralCommissioning.ArmFailSafe
2. NetworkCommissioning.AddOrUpdateWifiNetwork
3. NNetworkCommissioning.ConnectNetwork
4. DNS-SD -> success
5. [over IP] GeneralCommissioning.CommissioningComplete - disarm failsafe/CommitConfiguration

**Reconfiguration, device already in Wi-Fi network (failure)**
1. GeneralCommissioning.ArmFailSafe
2. NetworkCommissioning.RemoveNetwork
3. NetworkCommissioning.AddOrUpdateWifiNetwork(new SSID, new password)
4. NetworkCommissioning.ConnectNetwork
5. DNS-SD -> failure - RevertConfiguration - fallback to the last valid network

**Reconfiguration, device already in Wi-Fi network (success)** -> hard to test with `chip-tool` as it would require to configure 2 WiFi networks in parallel.

### DIFF
--- a/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
+++ b/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
@@ -147,6 +147,11 @@ CHIP_ERROR NrfWiFiDriver::CommitConfiguration()
 
 CHIP_ERROR NrfWiFiDriver::RevertConfiguration()
 {
+    if (WiFiManager::StationStatus::CONNECTING <= WiFiManager::Instance().GetStationStatus())
+    {
+        WiFiManager::Instance().Disconnect();
+    }
+
     LoadFromStorage();
 
     if (mStagingNetwork.IsConfigured())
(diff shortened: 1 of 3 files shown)


kkasperczyk-no (approved):
LGTM, let's discuss during our daily meeting whether we should include this commit in open upmerge or create a separate PR to sdk-nrf manifest to run tests on this one.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -147,6 +147,11 @@ CHIP_ERROR NrfWiFiDriver::CommitConfiguration()
 
 CHIP_ERROR NrfWiFiDriver::RevertConfiguration()
 {
+    if (WiFiManager::StationStatus::CONNECTED <= WiFiManager::Instance().GetStationStatus())
+    {
+        WiFiManager::Instance().Disconnect();
```
Damian-Nordic, line 152: What if the device is not yet connected but is connecting, i.e. wrong SSID has been provided? We should stop the connection establishment, right?

markaj-nordic: Ok, I checked if the disconnect request is sane when the supplicant is during connection establishment - and it should work fine, I will fix it.

markaj-nordic: Done

Damian-Nordic (approved):

PR #145, opened 2022-11-18
[nrf noup] Introduced a workaround to send Router Solicitation after …

kkasperczyk-no:
Router Solicitation is not sent after connecting to the Wi-Fi network by the Wi-Fi driver, so in result Thread Border Router doesn't send Router Advertisement to the device. As a workaround sending RS was added in the Matter platform code.

### DIFF
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -22,15 +22,17 @@
 
 #include "WiFiManager.h"
 
+#include <crypto/RandUtils.h>
 #include <inet/InetInterface.h>
 #include <inet/UDPEndPointImplSockets.h>
 #include <lib/support/logging/CHIPLogging.h>
 #include <platform/CHIPDeviceLayer.h>
 #include <platform/Zephyr/InetUtils.h>
 
-#include <zephyr/net/net_stats.h>
 #include <zephyr/kernel.h>
 #include <zephyr/net/net_event.h>
+#include <zephyr/net/net_if.h>
+#include <zephyr/net/net_stats.h>
 
 extern "C" {
 #include <common/defs.h>
@@ -359,6 +361,25 @@ void WiFiManager::ScanDoneHandler(uint8_t * data)
     }
 }
 
+void WiFiManager::SendRouterSolicitation(System::Layer * layer, void * param)
+{
+    net_if * iface = InetUtils::GetInterface();
+    if (iface && iface->if_dev->link_addr.type == NET_LINK_ETHERNET)
+    {
+        net_if_start_rs(iface);
+        Instance().mRouterSolicitationCounter++;
+        if (Instance().mRouterSolicitationCounter < kRouterSolicitationMaxCount)
+        {
+            DeviceLayer::SystemLayer().StartTimer(System::Clock::Milliseconds32(kRouterSolicitationIntervalMs),
+                                                  SendRouterSolicitation, nullptr);
+        }
+        else
+        {
+            Instance().mRouterSolicitationCounter = 0;
+        }
+    }
+}
+
 void WiFiManager::ConnectHandler(uint8_t * data)
 {
     const wifi_status * status      = reinterpret_cast<const wifi_status *>(data);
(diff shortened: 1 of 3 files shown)


kkasperczyk-no:
I did manual tests and it seems still this fix solves the problems with Router Solicitation. After applying it TBR prefix is assigned just after connecting to Wi-Fi. Removing DNM label then.

src/platform/nrfconnect/wifi/WiFiManager.h
```diff
@@ -164,6 +164,9 @@ class WiFiManager
         }
     };
 
+    static constexpr uint16_t kRouterSolicitationIntervalMs = 300;
```
Damian-Nordic, line 167: Based on https://www.rfc-editor.org/rfc/rfc4861#section-6.3.7, we shouldn't send RS more often than `RTR_SOLICITATION_INTERVAL = 4 seconds`. The initial RS delay after the connection establishment should be randomized between 0 and `MAX_RTR_SOLICITATION_DELAY  = 1 second`. Not sure how strict we need to be given that this is the workaround, but fyi that's how the host should behave.

Damian-Nordic (approved):

src/platform/nrfconnect/wifi/WiFiManager.h
kkasperczyk-no: I wasn't aware of such restrictions. It's not a big deal so I aligned it. It's always better to be compliant than not.

src/platform/nrfconnect/wifi/WiFiManager.h
```diff
@@ -193,6 +197,7 @@ class WiFiManager
     static void ConnectHandler(uint8_t * data);
     static void DisconnectHandler(uint8_t * data);
     static void PostConnectivityStatusChange(ConnectivityChange changeType);
+    static void _SendRouterSolicitation(System::Layer * layer, void * param);
```
markaj-nordic, line 200: nit: why using the underscore prefix?

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -359,6 +361,25 @@ void WiFiManager::ScanDoneHandler(uint8_t * data)
     }
 }
 
+void WiFiManager::_SendRouterSolicitation(System::Layer * layer, void * param)
+{
+    net_if * iface = InetUtils::GetInterface();
+    if (iface && iface->if_dev->link_addr.type == NET_LINK_ETHERNET)
+    {
+        net_if_start_rs(iface);
```
markaj-nordic, line 369: The docs of `net_if_start_rs` says: "Start neighbor discovery and send router solicitation message."
https://docs.zephyrproject.org/apidoc/latest/group__net__if.html#gab9bdb7f8a9eeed4d9b70965b8af82cb7

@kkasperczyk-no do you know what does it actually mean to "start neighbor discovery"? I mean, if we really need to start it every time we send the RS? Should' we call the `net_if_stop_rs` anywhere?

kkasperczyk-no: Actually no, looking at the implementation it just sends single RS packet through the given interface. Looking at the code Zephyr doesn't call net_if_stop_rs anywhere beside the case when it wants to cancel sending RS before it was done (at least that's my guess looking at the code).

src/platform/nrfconnect/wifi/WiFiManager.h
kkasperczyk-no: It was moved from the ConnectivityManagerImplWiFi that for some reason uses such convention for private methods. I removed the underscore.

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Ok, maybe the function name is just a bit misleading.

markaj-nordic (approved):

src/platform/nrfconnect/wifi/WiFiManager.cpp
Damian-Nordic: Hmm. Actually, the function looks like sending RS a few times. Otherwise, it's hard to explain this:
```
sys_slist_append(&active_rs_timers, &ipv6->rs_node);
```
Explicit stop is not needed because `rs_timeout` calls `sys_slist_remove(&active_rs_timers, NULL, &ipv6->rs_node);`. Maybe the impl is just broken? Did you see that it doesn't send multiple RSes or you assumed so @kkasperczyk-no?

Damian-Nordic: The impl seems incorrect in sending RS every second rather than 4 seconds, but this is a different issue.

kkasperczyk-no: I tested it some time ago and I saw on the wireshark, as many RS packets, as I called this method, so I assumed it is working fine.

LuDuda (approved):

PR #150, opened 2022-11-24
[nrf noup] Minor fixes for heap usage

Damian-Nordic:
* Fix build with CHIP_MALLOC_SYS_HEAP Kconfig option that replaces default malloc/free with alternatives based on Zephyr's sys_heap.
* If the default heap is used, error out if less than 16kB of RAM has been left for the heap.

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig.defaults
+++ b/config/nrfconnect/chip-module/Kconfig.defaults
@@ -85,6 +85,11 @@ config NEWLIB_LIBC_NANO
     bool
     default y
 
+# Warn a user if memory left for the heap is too small
+config NEWLIB_LIBC_MIN_REQUIRED_HEAP_SIZE
+    int
+    default 16384
+
 # Generic networking options
 config NET_SOCKETS_POSIX_NAMES
     bool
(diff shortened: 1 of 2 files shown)


ArekBalysNordic (approved):

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -85,6 +85,11 @@ config NEWLIB_LIBC_NANO
     bool
     default y
 
+# Warn a user if memory left for the heap is too small
+config NEWLIB_LIBC_MIN_REQUIRED_HEAP_SIZE
+    int
+    default 16384
```
markaj-nordic, line 91: How was this value determined?

markaj-nordic (approved):

config/nrfconnect/chip-module/Kconfig.defaults
Damian-Nordic: I remember that when testing HeapHighWatermark attribute (maximum heap usage so far) it's never reached 10kB. So 16kB looks quite safe to me. Or at least it's safer than the default 2kB. We can think about some automated tests to verify if that holds.

PR #151, opened 2022-11-25
[nrf noup] Decrease the minimum heap size for newlibc

markaj-nordic:
In some configurations (i.e. with BLE SMP enabled) we cannot achieve 16kB of free RAM for newlibc. 16kB was set with redundancy in mind, so we can make this requirement less strict (12kB).


LuDuda (approved):

Damian-Nordic (approved):

Damian-Nordic:
I am actually worried that on some samples we have this little memory left. Perhaps, we should tune the stack sizes etc :)

PR #152, opened 2022-11-25
[nrf noup] Wi-Fi status handling fixes

markaj-nordic:
* Increase the SYSTEM_WORKQUEUE_STACK_SIZE to match supplicant needs
* Decrease the connection timeout to be lower than failsafe (35s)
* Adapt WiFiRequestStatus to follow supplicant implementation
* All of above makes the failsafe more robust

Note that the following patch in wpa_supplicant is needed to have the proper connection timeout handling:
https://github.com/nrfconnect/sdk-hostap/pull/45

### DIFF
--- a/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
+++ b/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
@@ -105,7 +105,7 @@ CHIP_ERROR NrfWiFiDriver::Init(NetworkStatusChangeCallback * networkStatusChange
     {
         WiFiManager::ConnectionHandling handling{ [] { Instance().OnNetworkStatusChanged(Status::kSuccess); },
                                                   [] { Instance().OnNetworkStatusChanged(Status::kUnknownError); },
-                                                  System::Clock::Timeout{ kWiFiConnectNetworkTimeoutSeconds } };
+                                                  System::Clock::Seconds32{ kWiFiConnectNetworkTimeoutSeconds } };
         ReturnErrorOnFailure(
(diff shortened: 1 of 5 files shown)


kkasperczyk-no (approved):
LGTM, but please remember to always create a PR to sdk-nrf referenced to sdk-connectedhomeip PR as well. Otherwise sdk-nrf tests will not be triggered before we merge this to sdk-connectedhomeip master, what should not happen.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -105,7 +105,7 @@ CHIP_ERROR NrfWiFiDriver::Init(NetworkStatusChangeCallback * networkStatusChange
     {
         WiFiManager::ConnectionHandling handling{ [] { Instance().OnNetworkStatusChanged(Status::kSuccess); },
                                                   [] { Instance().OnNetworkStatusChanged(Status::kUnknownError); },
-                                                  System::Clock::Timeout{ kWiFiConnectNetworkTimeoutSeconds * 1000 } };
+                                                  System::Clock::Timeout{ kWiFiConnectNetworkTimeoutSeconds } };
```
Damian-Nordic, line 108: Why this is not needed given that `using Timeout = Milliseconds32;`?

Damian-Nordic (approved):

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
kkasperczyk-no: @markaj-nordic I pushed commit addressing this comment, as we need to move forward and merge this PR.

PR #170, opened 2022-12-20
[nrf noup] Fixed the timeout when connecting to the WPA3 secured APs

markaj-nordic:
Set the WiFi MFP (management frame protection) as at least optional. The current WiFi driver does not support this parameter within the scan result, so it must the hard coded for the time being (made it mandatory for secure associations, just in case).


Damian-Nordic (approved):

LuDuda (approved):

markaj-nordic:
@Damian-Nordic, @LuDuda I have also added a commit which tweaks the default nrf7002 revision.

LuDuda (approved):

PR #180, opened 2023-01-03
[nrf noup] [mrp] Added config to allow overriding MRP params at runtime

ArekBalysNordic:
Sometimes in different networks, it is needed to change MRP intervals at runtime. For example in low-power Wi-Fi networks when DTIM and beacon intervals may be various and depend on Access point settings.
The new config allows overriding MRP params at runtime by using the OverrideLocalMRPConfig() function, and clearing that override by using ClearLocalMRPConfigOverride().

### DIFF
--- a/src/platform/nrfconnect/BUILD.gn
+++ b/src/platform/nrfconnect/BUILD.gn
@@ -95,6 +95,8 @@ static_library("nrfconnect") {
       "wifi/WiFiManager.cpp",
       "wifi/WiFiManager.h",
     ]
+
+    public_deps += [ "${chip_root}/src/messaging:messaging_mrp_config" ]
   }
 
   if (chip_enable_nfc) {
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -29,6 +29,10 @@
 #include <platform/CHIPDeviceLayer.h>
 #include <platform/Zephyr/InetUtils.h>
 
+#if CHIP_CONFIG_ALLOW_RUNTIME_MRP_OVERRIDE
+#include <messaging/ReliableMessageProtocolConfig.h>
+#endif
+
 #include <zephyr/kernel.h>
 #include <zephyr/net/net_event.h>
 #include <zephyr/net/net_if.h>
@@ -45,10 +49,20 @@ extern "C" {
 extern char * net_sprint_ll_addr_buf(const uint8_t * ll, uint8_t ll_len, char * buf, int buflen);
 }
 
+// TODO When API for obtaining beacon interval and DTIM params is ready the following definitions can be removed
+#if CHIP_CONFIG_ALLOW_RUNTIME_MRP_OVERRIDE
+// The default value of Delivery Traffic Indication Message, a multiplier of the beacon interval.
+#define DEFAULT_DTIM_INTERVAL (3)
+// The default value of the beacon interval in miliseconds
+#define DEFAULT_BEACON_INTERVAL_MS (100_ms32)
+#endif
+
 namespace chip {
 namespace DeviceLayer {
 
 namespace {
+
+using namespace chip::System::Clock::Literals;
 
 NetworkCommissioning::WiFiScanResponse ToScanResponse(const wifi_scan_result * result)
 {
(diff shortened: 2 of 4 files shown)


src/platform/nrfconnect/BUILD.gn
```diff
@@ -95,6 +95,8 @@ static_library("nrfconnect") {
       "wifi/WiFiManager.cpp",
       "wifi/WiFiManager.h",
     ]
+
+    public_deps += [ "${chip_root}/src/messaging:messaging_mrp_config" ]
```
kkasperczyk-no, line 99: Maybe let's move it under the `chip_enable_wifi` brackets, as it doesn't seem to be needed for Thread.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -26,6 +26,7 @@
 #include <inet/InetInterface.h>
 #include <inet/UDPEndPointImplSockets.h>
 #include <lib/support/logging/CHIPLogging.h>
+#include <messaging/ReliableMessageProtocolConfig.h>
```
kkasperczyk-no, line 29: Add #if CHIP_CONFIG_ALLOW_RUNTIME_MRP_OVERRIDE.

kkasperczyk-no (commented):
Could you also post a PR to upstream? Ideally we should first push the changes to upstream and try to pull it here, as fromtree or at least fromlist. I know we sometimes do that in a different way, but here it doesn't seem to be needed.

src/platform/nrfconnect/BUILD.gn
ArekBalysNordic: Hmm, it is there already. Please check it one more time.

ArekBalysNordic:
> Could you also post a PR to upstream? Ideally we should first push the changes to upstream and try to pull it here, as fromtree or at least fromlist. I know we sometimes do that in a different way, but here it doesn't seem to be needed.

Sure, that's the plan. But first, I want to know that it has been done in a proper way. :)

src/platform/nrfconnect/BUILD.gn
kkasperczyk-no: Oh right, that's my oversight.

LuDuda (changes requested):
I agree, we need to push upstream first and the do cherry-pick instead.

Damian-Nordic (commented):
I think it would be cleaner to do it by setting `CHIP_DEVICE_CONFIG_ENABLE_SED` for Wi-Fi power saving mode and implementing `ConnectivityManager::GetSEDIntervalsConfig`. It would more consistent with Thread SEDs and give you two more extra features:
- MRP backoff adjustment that takes into account sender's sleepy active interval (see `ReliableMessageMgr::GetBackoff`)
- Ability to emit `kSEDIntervalChange` event when beacon/DTIM intervals change in order to refresh advertised SAI/SII parameters.

All you need to do is:
- set `CHIP_DEVICE_CONFIG_ENABLE_SED` when Wi-Fi low-power mode is enabled
- Add these methods in nrfconnect `ConnectivityManagerImplWiFi`:
  ```
    CHIP_ERROR _GetSEDIntervalsConfig(ConnectivityManager::SEDIntervalsConfig & intervalsConfig); // do a proper impl that returns DTIM * beacon interval
    CHIP_ERROR _SetSEDIntervalsConfig(const ConnectivityManager::SEDIntervalsConfig & intervalsConfig) { return CHIP_ERROR_UNSUPPORTED_FEATURE; }
    CHIP_ERROR _RequestSEDActiveMode(bool onOff) { return CHIP_NO_ERROR; }
    ```
- Add the following lines to nrfconnect `ConnectivityManager.h` to resolve conflict between `ConnectivityManager_Wifi` and `ConnectivityManager_NoThread`:
  ```
  using ConnectivityManager_Wifi::_GetSEDIntervalsConfig;
  using ConnectivityManager_Wifi::_SetSEDIntervalsConfig;
  using ConnectivityManager_Wifi::_RequestSEDActiveMode;
  ```

The drawback is that it will enable the logic for switching between active and idle mode which for now won't be supported in our Wi-Fi implementation, but we can simply make it a no-op, or add another config for that specific feature to save some flash (if it's worth the effort).

What are your thoughts?

LuDuda:
> I think it would be cleaner to do it by setting `CHIP_DEVICE_CONFIG_ENABLE_SED` for Wi-Fi power saving mode and implementing `ConnectivityManager::GetSEDIntervalsConfig`. It would more consistent with Thread SEDs and give you two more extra features:
> 
> * MRP backoff adjustment that takes into account sender's sleepy active interval (see `ReliableMessageMgr::GetBackoff`)
> * Ability to emit `kSEDIntervalChange` event when beacon/DTIM intervals change in order to refresh advertised SAI/SII parameters.
> 
> All you need to do is:
> 
> * set `CHIP_DEVICE_CONFIG_ENABLE_SED` when Wi-Fi low-power mode is enabled
> * Add these methods in nrfconnect `ConnectivityManagerImplWiFi`:
>   ```
>     CHIP_ERROR _GetSEDIntervalsConfig(ConnectivityManager::SEDIntervalsConfig & intervalsConfig); // do a proper impl that returns DTIM * beacon interval
>     CHIP_ERROR _SetSEDIntervalsConfig(const ConnectivityManager::SEDIntervalsConfig & intervalsConfig) { return CHIP_ERROR_UNSUPPORTED_FEATURE; }
>     CHIP_ERROR _RequestSEDActiveMode(bool onOff) { return CHIP_NO_ERROR; }
>   ```
> * Add the following lines to nrfconnect `ConnectivityManager.h` to resolve conflict between `ConnectivityManager_Wifi` and `ConnectivityManager_NoThread`:
>   ```
>   using ConnectivityManager_Wifi::_GetSEDIntervalsConfig;
>   using ConnectivityManager_Wifi::_SetSEDIntervalsConfig;
>   using ConnectivityManager_Wifi::_RequestSEDActiveMode;
>   ```
> 
> The drawback is that it will enable the logic for switching between active and idle mode which for now won't be supported in our Wi-Fi implementation, but we can simply make it a no-op, or add another config for that specific feature to save some flash (if it's worth the effort).
> 
> What are your thoughts?

Great idea!

I know it's a bit off.. But would it be easier to change SED to ICD - would have more sense for Wi-Fi..

Damian-Nordic:
> I know it's a bit off.. But would it be better to rename SED to ICD in the API/KConfigs - would have more sense for Wi-Fi.. In the other hand it will break compatibility.

It's more complicated :) Those parameters are still advertised as SAI (**Sleepy** Active Interval) so I'm not sure what would be the best name.

ArekBalysNordic:
Yes, I'm closing it.

PR #181, opened 2023-01-04
[nrf noup] Increased heap size for Wi-Fi.

kkasperczyk-no:
Empirically it was measured that heap size needed to pass the Wi-Fi commissioning is ~25k. To be on the safe side let's set it to 28k.


kkasperczyk-no:
> I presume we are going to keep the statically allocated heap for all samples, including Wi-Fi, and the mDNS fix will save us from hard faults if the heap is still too small? 

Yes, mDNS fix will just allow to crash the application and print error message instead of a hard fault. 

> @kkasperczyk-no I am also curious how this was measured. Won't this affect the overall stack capacity/capability?

I used dynamically allocated heap size to see what is the real usage of it. In the Zephyr's newlibc hooks implementation there is one called `_sbrk`. It is called every time system needs to allocate new chunk of memory, so it is possible to check the heap offset compared to begin after commissioning and estimate the size.

> Won't this affect the overall stack capacity/capability?

I don't think we have much choice. We need to let application allocate this memory for heap or it will just won't work. The difference is that statically allocated memory is not freed to the system, so in theory it sometimes may be wasted, but on the other hand it reduces RAM size usage and we can predict the max usage. Both solutions have some strong and weak points, so I believe we should give it a chance and see what CI results will show as after some time.

markaj-nordic:
> > I presume we are going to keep the statically allocated heap for all samples, including Wi-Fi, and the mDNS fix will save us from hard faults if the heap is still too small?
> 
> Yes, mDNS fix will just allow to crash the application in controlled way and print error message instead of a hard fault.
> 
> > @kkasperczyk-no I am also curious how this was measured.
> 
> I used dynamically allocated heap size to see what is the real usage of it. In the Zephyr's newlibc hooks implementation there is one called `_sbrk`. It is called every time system needs to allocate new chunk of memory, so it is possible to check the heap offset compared to begin after commissioning and estimate the size.
> 
> > Won't this affect the overall stack capacity/capability?
> 
> I don't think we have much choice. We need to let application allocate this memory for heap or it will just won't work. The difference is that statically allocated memory is not freed to the system, so in theory it sometimes may be wasted, but on the other hand it reduces RAM size usage and we can predict the max usage. Both solutions have some strong and weak points, so I believe we should give it a chance and see what CI results will show as after some time.

Yehh, of course. I am just a bit concerned because this rather a significant increase. Thanks for the explanation.

Damian-Nordic (approved):

PR #182, opened 2023-01-09
[nrf noup] Added support for multi-variant application OTA DFU

markaj-nordic:
* Implemented dedicated flash write handlers for a Thread/WiFi switchable app.
* A script for generating multi variant OTA image has been added in sdk-nrf.

### DIFF
--- a/config/nrfconnect/chip-module/CMakeLists.txt
+++ b/config/nrfconnect/chip-module/CMakeLists.txt
@@ -222,6 +222,7 @@ chip_gn_arg_bool  ("chip_detail_logging"                    CONFIG_MATTER_LOG_LE
 chip_gn_arg_bool  ("chip_automation_logging"                "false")
 chip_gn_arg_bool  ("chip_malloc_sys_heap"                   CONFIG_CHIP_MALLOC_SYS_HEAP)
 chip_gn_arg_bool  ("chip_enable_wifi"                       CONFIG_WIFI_NRF700X)
+chip_gn_arg_bool  ("chip_multi_variant_ota"                 CONFIG_THREAD_WIFI_SWITCHING)
 
 if (CONFIG_CHIP_FACTORY_DATA)
     chip_gn_arg_bool  ("chip_use_transitional_commissionable_data_provider"  "false")
--- a/src/platform/nrfconnect/OTAImageProcessorImpl.cpp
+++ b/src/platform/nrfconnect/OTAImageProcessorImpl.cpp
@@ -38,6 +38,10 @@
 #include <zephyr/logging/log.h>
 #include <zephyr/pm/device.h>
 
+#ifdef CONFIG_THREAD_WIFI_SWITCHING
+#include "OTAMultiImageDownloader.h"
+#endif
+
 #if CONFIG_CHIP_CERTIFICATION_DECLARATION_STORAGE
 // Cd globals are needed to be accessed from dfu image writer lambdas
 namespace {
@@ -48,6 +52,25 @@ size_t sCdSavedBytes                                      = 0;
 
 namespace chip {
 namespace DeviceLayer {
+
+static int DefaultWriterInit(int id, size_t size)
+{
+    return dfu_target_init(DFU_TARGET_IMAGE_TYPE_MCUBOOT, id, size, nullptr);
+}
+
+static int DefaultWriterWrite(const uint8_t * chunk, size_t chunk_size)
+{
+    return dfu_target_write(chunk, chunk_size);
+}
+
+static int DefaultWriterClose(bool success)
+{
+    return success ? dfu_target_done(success) : dfu_target_reset();
+}
+
+static dfu_image_open_t sImageOpenHandler   = DefaultWriterInit;
+static dfu_image_write_t sImageWriteHandler = DefaultWriterWrite;
+static dfu_image_close_t sImageCloseHandler = DefaultWriterClose;
 
 CHIP_ERROR OTAImageProcessorImpl::PrepareDownload()
 {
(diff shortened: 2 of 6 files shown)


src/platform/nrfconnect/OTAMultiImageDownloader.cpp
```diff
@@ -0,0 +1,181 @@
...
+}
+
+int OTAMultiImageDownloader::Finalize()
+{
+    int err = stream_flash_buffered_write(&mStream, nullptr, 0, true);
+    k_free(mBuffer);
+    return err;
+}
+
+// Implementation of stateless OTAMultiImageDownloaders class
```
markaj-nordic, line 104: Maybe this can be replaced with just a namespace.

src/platform/nrfconnect/OTAImageProcessorImpl.cpp
```diff
@@ -114,6 +144,10 @@ CHIP_ERROR OTAImageProcessorImpl::Apply()
     // Schedule update of all images
     int err = dfu_target_schedule_update(-1);
 
+#ifdef CONFIG_THREAD_WIFI_SWITCHING
+    err = OTAMultiImageDownloaders::Apply();
```
Damian-Nordic, line 148: Shouldn't it be rather called in `ConfirmCurrentImage`? That is, when we know that the image has been applied correctly and we already execute it successfully?

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -173,6 +173,11 @@ config NFC_THREAD_CALLBACK
 config CHIP_OTA_REQUESTOR
     default y
 
+if THREAD_WIFI_SWITCHING
```
Damian-Nordic, line 176: This is defined just in the lock sample, right? If so, it may be better to modify `DFU_MULTI_IMAGE_MAX_IMAGE_COUNT` there?

src/platform/nrfconnect/OTAMultiImageDownloader.cpp
```diff
@@ -0,0 +1,181 @@
...
+    {
+        ChipLogError(SoftwareUpdate, "Flash device is not ready!");
+        return -EFAULT;
+    }
+    return Init(flash, mOffset, size);
+}
+
+int OTAMultiImageDownloader::Init(const device * flash, size_t offset, size_t size)
+{
+    mBuffer = (uint8_t *) k_malloc(kBufferSize);
```
Damian-Nordic, line 53: I think using kernel heap may not be the best choice given how precious resource there is. We should better use `Platform::MemoryAlloc` or even `ScopedBuffer` for safer memory management.

src/platform/nrfconnect/OTAMultiImageDownloader.cpp
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    OTAMultiImageDownloaders::SetCurrentImageId(id);
+    return OTAMultiImageDownloaders::GetCurrentImage()->Init(sExtFlash, size);
+}
+
+int OTAMultiImageDownloaders::Write(const uint8_t * chunk, size_t chunk_size)
+{
+    if (OTAMultiImageDownloaders::CurrentImageId() <= CurrentImageID::APP_1_CORE_NET)
```
Damian-Nordic, line 140: I struggle to understand this condition. What if we're currently running a Wi-Fi variant?

Damian-Nordic (commented):
A few initial comments. I wonder if it wouldn't be cleaner if you created another `MultiVariantOTAImageProcessor` class that inherits from `OTAImageProcessorImpl`, overrides needed methods and keeps all the necessary state in itself. That way you could avoid mixing instance-based code in `OTAImageProcessorImpl` with `OTAMultiImageDownloaders` that uses global state. However, I admit we have plenty of code that uses global state right now, so I'm fine with the current approach, either :)

src/platform/nrfconnect/OTAImageProcessorImpl.cpp
markaj-nordic: Ahh sure, my bad.

src/platform/nrfconnect/OTAMultiImageDownloader.cpp
markaj-nordic: Ok, will do.

config/nrfconnect/chip-module/Kconfig.defaults
markaj-nordic: Ok, I will move it to the lock sample.

src/platform/nrfconnect/OTAMultiImageDownloader.cpp
markaj-nordic: Yep, look like a bug. Will fix it.

markaj-nordic:
> A few initial comments. I wonder if it wouldn't be cleaner if you created another `MultiVariantOTAImageProcessor` class that inherits from `OTAImageProcessorImpl`, overrides needed methods and keeps all the necessary state in itself. That way you could avoid mixing instance-based code in `OTAImageProcessorImpl` with `OTAMultiImageDownloaders` that uses global state. However, I admit we have plenty of code that uses global state right now, so I'm fine with the current approach, either :)

I get your point. Let me see if I can embrace this approach. For now I have addressed all comments regarding current implementation.

markaj-nordic:
> A few initial comments. I wonder if it wouldn't be cleaner if you created another `MultiVariantOTAImageProcessor` class that inherits from `OTAImageProcessorImpl`, overrides needed methods and keeps all the necessary state in itself. That way you could avoid mixing instance-based code in `OTAImageProcessorImpl` with `OTAMultiImageDownloaders` that uses global state. However, I admit we have plenty of code that uses global state right now, so I'm fine with the current approach, either :)

I get your point. Let me see if I can embrace this approach. For now I have addressed all comments regarding current implementation.

> > A few initial comments. I wonder if it wouldn't be cleaner if you created another `MultiVariantOTAImageProcessor` class that inherits from `OTAImageProcessorImpl`, overrides needed methods and keeps all the necessary state in itself. That way you could avoid mixing instance-based code in `OTAImageProcessorImpl` with `OTAMultiImageDownloaders` that uses global state. However, I admit we have plenty of code that uses global state right now, so I'm fine with the current approach, either :)
> 
> I get your point. Let me see if I can embrace this approach. For now I have addressed all comments regarding current implementation.

The problem I see with this approach is that we need to register dfu_multi_image callbacks which by definition must be implemented as stateless (i.e. we cannot capture any non static member of OTAImageProcessorImpl). Yet we still have to track the currently processed image ID to decide which external flash slots must be targeted for a given image chunk. So it looks like we cannot do without a globally scoped state (current image ID). It seems to me that it could be resolved by modification of dfu_multi_image API by either providing current image ID in all callbacks (as it is done only for open callback currently) or (maybe) by allowing a user to register multiple writers for given ID. Regardless of that we can still inherit from OTAImageProcessorImpl but I am not sure if this brings an added value (well it's a tradeoff between having more neat code/more files in the repo).

config/nrfconnect/chip-module/CMakeLists.txt
```diff
@@ -222,6 +222,7 @@ chip_gn_arg_bool  ("chip_detail_logging"                    CONFIG_MATTER_LOG_LE
 chip_gn_arg_bool  ("chip_automation_logging"                "false")
 chip_gn_arg_bool  ("chip_malloc_sys_heap"                   CONFIG_CHIP_MALLOC_SYS_HEAP)
 chip_gn_arg_bool  ("chip_enable_wifi"                       CONFIG_WIFI_NRF700X)
+chip_gn_arg_bool  ("chip_multi_variant_ota"                 CONFIG_THREAD_WIFI_SWITCHING)
```
kkasperczyk-no, line 225: That's a minor, but maybe it would be cleaner to create a dedicated Kconfig? I would imagine it could be CONFIG_MULTI_VARIANT_OTA and then imply it in the CONFIG_THREAD_WIFI_SWITCHING.

markaj-nordic: IMHO the clean solution would be simply to support OTA whenever the CHIP_OTA_REQUESTOR is set, regardless if we have single or multiple variant application. This is actually the current state. On the other hand I agree that mapping the CONFIG_THREAD_WIFI_SWITCHING into chip_multi_variant_ota does not look clean... but @kkasperczyk-no do you have any idea how to make it more neat without introducing another Kconfig (which preferably I would like to avoid)? I mean we may just build all OTA related files but that's rather a no-go considering our flash limitations.

kkasperczyk-no: But why would you like to avoid it? I know it may seem we have already bunch of kconfig options, but the fact is that our platform is very complex and this is the most effective way of configuring our implementation to disable certain parts of a code. 
Honestly I thought this functionality will be used only by the switched sample, so we will somehow implement it in the sample code to avoid making our platform implementation even more complex with another variant of it. Nevertheless I understand it is much more convenient to do that in the platform. In such case I would recommend wrap it into new kconfig option to make some new functionality clearly separated from the single sample.

markaj-nordic:
After f2f discussion the decision is to move these implementation to nrfconnect SDK. Closing this PR.

PR #190, opened 2023-01-19
[nrf fromtree][icd] Removed ICD related methods (former SED) from NoT…

kkasperczyk-no:
…hreadImpl (#24444)

There are several methods related to ICD (former SED) used for setting expected communication intervals. Currently they are implemented only for Thread devices and for Wi-Fi the implementation returns unsupported feature through NoThreadImpl. Including this sleep intervals could be useful for some low power Wi-Fi modes too.

Changes:
* Removed methods implementation from NoThread impl. Now after enabling CHIP_DEVICE_CONFIG_ENABLE_SED (I believe it should be renamed to ICD in the future), the implementation will have to be provided by the proper platform.
* Implemented methods for nrfconnect Wi-Fi platform.

(cherry picked from commit f234b94403d682241ea18ab6e9902c9094022a22)


examples/light-switch-app/nrfconnect/Kconfig
```diff
@@ -40,6 +40,13 @@ config CHIP_ENABLE_SLEEPY_END_DEVICE_SUPPORT
 
 endif # NET_L2_OPENTHREAD
 
+if CHIP_WIFI
+
+config NRF_WIFI_LOW_POWER
```
markaj-nordic, line 45: Why not enabling the low power mode for all our samples, i.e. by removing the NRF_WIFI_LOW_POWER=n from Kconfig.defaults? And what about NCS samples?

markaj-nordic (approved):

ArekBalysNordic (approved):

examples/light-switch-app/nrfconnect/Kconfig
kkasperczyk-no: > Why not enabling the low power mode for all our samples, i.e. by removing the NRF_WIFI_LOW_POWER=n from Kconfig.defaults?

Similarly to Thread, because we do not always need it. It only has sense for battery supplied devices. For devices like light bulb enabling low power mode and thus reducing device performance is not reasonable.

> And what about NCS samples?

NCS samples have this option enabled in their Kconfig files. It was introduced in the corresponding sdk-nrf PR: https://github.com/nrfconnect/sdk-nrf/pull/9978

markaj-nordic: Ahh ok, I haven't noticed the ncs PR. Thanks.

PR #192, opened 2023-01-19
[nrf fromlist] Increase MRP intervals for Thread

Damian-Nordic:
Increase the default MRP parameters for Thread devices from 300 to 800ms.

### DIFF
--- a/src/messaging/ReliableMessageProtocolConfig.h
+++ b/src/messaging/ReliableMessageProtocolConfig.h
@@ -34,31 +34,54 @@ namespace chip {
 /**
  *  @def CHIP_CONFIG_MRP_LOCAL_ACTIVE_RETRY_INTERVAL
  *
- *  @brief
- *    Active retransmit interval, or time to wait before retransmission after
- *    subsequent failures in milliseconds.
+ *  @brief Base retry interval of the present node when it is in the active state.
  *
- *  This is the default value, that might be adjusted by end device depending on its
- *  needs (e.g. sleeping period) using Service Discovery TXT record SAI key.
+ *  Base interval that a peer node should use to calculate the retransmission
+ *  timeout when it sends a message to the present node and the present node is
+ *  perceived by the peer as active.
  *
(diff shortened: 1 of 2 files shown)


src/messaging/ReliableMessageProtocolConfig.h
```diff
@@ -121,6 +144,27 @@ namespace chip {
...
+ *  rertansmissions in the scenario in which a sender that operates in a high-
+ *  latency network (such as Thread) sends a message to a receiver that operates
+ *  in a low-latency network (such as Wi-Fi). In this scenario, the SAI and SII
+ *  parameters advertised by the receiver are low although the average round-
+ *  trip time of a big request-response pair is long due to the nature of the
+ *  sender.
+ */
+#ifndef CHIP_CONFIG_MRP_RETRY_INTERVAL_SENDER_BOOST
+#if CHIP_ENABLE_OPENTHREAD
+#define CHIP_CONFIG_MRP_RETRY_INTERVAL_SENDER_BOOST (500_ms)
```
LuDuda, line 164: Just a question.. "ms" or "ms32" as above?

LuDuda: Do i correctly understand that 500ms was calculated as 800ms - 300ms?

LuDuda (commented):
Shouldn't we contribute it anyway? But merge here even without consensus in upstream?

LuDuda (approved):

src/messaging/ReliableMessageProtocolConfig.h
Damian-Nordic: Right, should be `_ms32`. It is cast, anyway, but that is more correct. Yeah, I wanted to make sure that the retransmission timeout is not lower than 800ms.

I'll contribute the changes to master, but our upstream tests are not functional at the moment so I need some confirmation that these values to trigger any other issues (which I can only get from NCS CI).

ArekBalysNordic (approved):

PR #193, opened 2023-01-20
[nrf noup] [nrfconnect] Introduce BLE advertising arbiter

Damian-Nordic:
Add a class that enables easier coexistence of Matter and custom BLE services.

### DIFF
--- a/src/platform/Zephyr/BLEManagerImpl.cpp
+++ b/src/platform/Zephyr/BLEManagerImpl.cpp
@@ -43,6 +43,8 @@
 #include <zephyr/sys/byteorder.h>
 #include <zephyr/sys/util.h>
 
+#include <array>
+
 using namespace ::chip;
 using namespace ::chip::Ble;
 using namespace ::chip::System;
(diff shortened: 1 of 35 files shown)


src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -233,35 +238,57 @@ struct BLEManagerImpl::ServiceData
     ChipBLEDeviceIdentificationInfo deviceIdInfo;
 } __attribute__((packed));
 
-CHIP_ERROR BLEManagerImpl::StartAdvertising(void)
+inline CHIP_ERROR BLEManagerImpl::PrepareAdvertisingRequest()
```
markaj-nordic, line 241: Why `inline` here?

markaj-nordic (approved):

src/platform/Zephyr/BLEManagerImpl.cpp
Damian-Nordic: This is a way to tell the compiler to use internal linkage for a method. I hope it can help compiler make the decision to inline this function (didn't check it though).

PR #194, opened 2023-01-23
Wifi ble coex

kkasperczyk-no:
Pulled commits to improve Bluetooth LE connection configuration, enable Wi-Fi/BLE coex and enable erasing settings partition using MCUMGR command during DFU over Bluetooth LE SMP.

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig.defaults
+++ b/config/nrfconnect/chip-module/Kconfig.defaults
@@ -108,8 +108,18 @@ config BT_PERIPHERAL_PREF_MIN_INT
 config BT_PERIPHERAL_PREF_MAX_INT
     default 36
 
+# Increase BT timeout to 5 s to improve connection reliability and avoid fast drop outs.
+config BT_PERIPHERAL_PREF_TIMEOUT
+    default 500
+
 config BT_GAP_AUTO_UPDATE_CONN_PARAMS
     default y
+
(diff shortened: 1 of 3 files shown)


config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -108,9 +108,19 @@ config BT_PERIPHERAL_PREF_MIN_INT
 config BT_PERIPHERAL_PREF_MAX_INT
     default 36
 
+# Increase BT timeout to 5 s to improve connection reliability and avoid fast drop outs.
```
ArekBalysNordic, line 111: Nit: Maybe it's worth adding an explanation why the BT timeout was increased to 5 s while the default value of the config is set to 500.

ArekBalysNordic (approved):

config/nrfconnect/chip-module/Kconfig.defaults
kkasperczyk-no: That's explained in this Kconfig definition, so I didn't want to copy the help text 
"config BT_PERIPHERAL_PREF_TIMEOUT
	int "Peripheral preferred supervision timeout in 10ms units""

Nevertheless I can add it, but it would require updating upstream first, so I can do that by the way of doing some other changes.

markaj-nordic (approved):

PR #198, opened 2023-01-25
[nrf noup] wifi: scan refinements

markaj-nordic:
* Currently using the specific channel number does not work reliably in WiFi driver, so use the generic channel screening.
*  In the case there are many networks available, the scan result events may saturate the CHIP work queue which has lower priority then the net_mgmt and main threads. So, collect the scan results in the main thread and only ping the CHIP thread when the scan is done.

### DIFF
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -336,7 +336,7 @@ void WiFiManager::ScanResultHandler(uint8_t * data)
             }
 
             Instance().mWiFiParams.mParams.timeout = Instance().mHandling.mConnectionTimeout.count();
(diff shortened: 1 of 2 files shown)


Damian-Nordic (approved):

markaj-nordic:
@Damian-Nordic I added another commit which aims to offload the CHIP work queue with scan result event. So please review again.

Damian-Nordic (approved):

LuDuda (approved):

LuDuda:
Would be good to contribute it as well at some point in time.

markaj-nordic:
> Would be good to contribute it as well at some point in time.

After discussing it f2f, the decision is to defer the contribution until we have a stable CI.

PR #203, opened 2023-02-03
[nrf noup] [nrfconnect] Introduce the WiFi connection recovery feature

ArekBalysNordic:
The WiFi connection recovery feature allows device re-scanning and re-connecting to the known WiFi network after the device's reboot and when the known SSID has not been found during the last scan. The connection recovery interval is doubled with
every occurrence to the defined maximum value and then its value depends on the maximum value +- the defined random jitter.

After restoring the connection, the Connection Recovery Interval is restored to the defined minimum value after elapsing of the defined delay to avoid frequent reconnections to a poor link quality network.

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig
+++ b/config/nrfconnect/chip-module/Kconfig
@@ -172,6 +172,51 @@ config CHIP_FACTORY_RESET_ERASE_NVS
 	bool
 	default y if CHIP_FACTORY_DATA || CHIP_FACTORY_DATA_CUSTOM_BACKEND
 
+config CHIP_WIFI_CONNECTION_RECOVERY_MINIMUM_INTERVAL
+	int "Define the minimum connection recovery time interval in milliseconds"
+	depends on CHIP_WIFI
+	default 500
+	help
+	  Specifies the minimum connection recovery interval (in milliseconds).
+
+config CHIP_WIFI_CONNECTION_RECOVERY_MAXIMUM_INTERVAL
+	int "Define the maximum connection recovery time interval in milliseconds"
+	depends on CHIP_WIFI
+	default 3600000 # 1 hour
+	help
+	  Specifies the maximum connection recovery interval (in milliseconds).
+
+config CHIP_WIFI_CONNECTION_RECOVERY_MAX_RETRIES_NUMBER
+	int "Define maximum amount of connection recovery occurrences"
+	depends on CHIP_WIFI
+	default 0
+	help
+	  Specifies the maximum connection recovery attempts. 
+	  If set to 0, no limitation is applied and the connect retry is performed endlessly.
+
+config CHIP_WIFI_CONNECTION_RECOVERY_JITTER
+	int "Define the connection recovery jitter in milliseconds"
+	depends on CHIP_WIFI
+	default 2000
+	help
+	  Specifies the maximum connection recovery jitter interval (in milliseconds).
+	  Once the wait time reaches the maximum value a random jitter interval is
+	  added to it to avoid periodicity. The random jitter is selected
+	  within range [-JITTER; +JITTER]. The random jitter value is added only
+	  when connection recovery interval reaches the maximum value.
+
+config CHIP_WIFI_CONNECTION_RECOVERY_RESET_DELAY
+	int "Define the delay between resetting connection recovery interval"
+	depends on CHIP_WIFI
+	default 0
+	help
+	  Specifies the delay after a successful connection recovery.
+	  The 0 value means that the connection recovery interval will be cleared
+	  immediately after the connection recovery.
+	  A non-zero value delays the interval reset, which is helpful in the situation
+	  where a link quality to a WiFi access point is poor and device disconnects 
+	  frequently.
+
 endif
 
 config CHIP_LOG_SIZE_OPTIMIZATION
--- a/config/nrfconnect/chip-module/Kconfig
+++ b/config/nrfconnect/chip-module/Kconfig
@@ -172,51 +172,6 @@ config CHIP_FACTORY_RESET_ERASE_NVS
 	bool
 	default y if CHIP_FACTORY_DATA || CHIP_FACTORY_DATA_CUSTOM_BACKEND
 
-config CHIP_WIFI_CONNECTION_RECOVERY_MINIMUM_INTERVAL
-	int "Define the minimum connection recovery time interval in milliseconds"
-	depends on CHIP_WIFI
-	default 500
-	help
-	  Specifies the minimum connection recovery interval (in milliseconds).
-
-config CHIP_WIFI_CONNECTION_RECOVERY_MAXIMUM_INTERVAL
-	int "Define the maximum connection recovery time interval in milliseconds"
-	depends on CHIP_WIFI
-	default 3600000 # 1 hour
-	help
-	  Specifies the maximum connection recovery interval (in milliseconds).
-
-config CHIP_WIFI_CONNECTION_RECOVERY_MAX_RETRIES_NUMBER
-	int "Define maximum amount of connection recovery occurrences"
-	depends on CHIP_WIFI
-	default 0
-	help
-	  Specifies the maximum connection recovery attempts. 
-	  If set to 0, no limitation is applied and the connect retry is performed endlessly.
-
-config CHIP_WIFI_CONNECTION_RECOVERY_JITTER
-	int "Define the connection recovery jitter in milliseconds"
-	depends on CHIP_WIFI
-	default 2000
-	help
-	  Specifies the maximum connection recovery jitter interval (in milliseconds).
-	  Once the wait time reaches the maximum value a random jitter interval is
-	  added to it to avoid periodicity. The random jitter is selected
-	  within range [-JITTER; +JITTER]. The random jitter value is added only
-	  when connection recovery interval reaches the maximum value.
-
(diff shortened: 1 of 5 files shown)


markaj-nordic:
@wiba-nordic could you please take a look at the KConfig descriptions?

config/nrfconnect/chip-module/Kconfig
```diff
@@ -172,6 +172,51 @@ config CHIP_FACTORY_RESET_ERASE_NVS
...
+
+config CHIP_WIFI_CONNECTION_RECOVERY_MAXIMUM_INTERVAL
+	int "Define the maximum connection recovery time interval in milliseconds"
+	depends on CHIP_WIFI
+	default 3600000 # 1 hour
+	help
+	  Specifies the maximum connection recovery interval (in milliseconds).
+
+config CHIP_WIFI_CONNECTION_RECOVERY_MAX_RETRIES_NUMBER
+	int "Define maximum amount of connection recovery occurrences"
```
wiba-nordic, line 190:
```suggestion
	int "Define the maximum amount of connection recovery occurrences"
```

config/nrfconnect/chip-module/Kconfig
```diff
@@ -172,6 +172,51 @@ config CHIP_FACTORY_RESET_ERASE_NVS
...
+	depends on CHIP_WIFI
+	default 0
+	help
+	  Specifies the maximum connection recovery attempts.
```
wiba-nordic, line 194:
```suggestion
	  Specifies the maximum number of connection recovery attempts. 
```

config/nrfconnect/chip-module/Kconfig
```diff
@@ -172,6 +172,51 @@ config CHIP_FACTORY_RESET_ERASE_NVS
...
+	help
+	  Specifies the maximum connection recovery interval (in milliseconds).
+
+config CHIP_WIFI_CONNECTION_RECOVERY_MAX_RETRIES_NUMBER
+	int "Define maximum amount of connection recovery occurrences"
+	depends on CHIP_WIFI
+	default 0
+	help
+	  Specifies the maximum connection recovery attempts. 
+	  If set to 0, no limitation is applied and the connect retry is performed endlessly.
```
wiba-nordic, line 195:
```suggestion
	  If set to 0, no limitation is applied and the connection attempts to recover indefinitely.
```

config/nrfconnect/chip-module/Kconfig
```diff
@@ -172,6 +172,51 @@ config CHIP_FACTORY_RESET_ERASE_NVS
...
+
+config CHIP_WIFI_CONNECTION_RECOVERY_JITTER
+	int "Define the connection recovery jitter in milliseconds"
+	depends on CHIP_WIFI
+	default 2000
+	help
+	  Specifies the maximum connection recovery jitter interval (in milliseconds).
+	  Once the wait time reaches the maximum value a random jitter interval is
```
wiba-nordic, line 203:
```suggestion
	  Once the wait time reaches the current maximum value, a random jitter interval is
```

config/nrfconnect/chip-module/Kconfig
```diff
@@ -172,6 +172,51 @@ config CHIP_FACTORY_RESET_ERASE_NVS
...
+	  added to it to avoid periodicity. The random jitter is selected
+	  within range [-JITTER; +JITTER]. The random jitter value is added only
+	  when connection recovery interval reaches the maximum value.
```
wiba-nordic, line 206: "JITTER" is is referring to current config "CHIP_WIFI_CONNECTION_RECOVERY_JITTER", right? If it is referring to some other value, it needs clarification, but if it is referring to its own value, it is clear enough.

Can you explain the last sentence to me? It seems to be the same information as the first sentence.

config/nrfconnect/chip-module/Kconfig
```diff
@@ -172,6 +172,51 @@ config CHIP_FACTORY_RESET_ERASE_NVS
...
+
+config CHIP_WIFI_CONNECTION_RECOVERY_RESET_DELAY
+	int "Define the delay between resetting connection recovery interval"
```
wiba-nordic, line 209:
```suggestion
	int "Define the delay before resetting connection recovery interval"
```

config/nrfconnect/chip-module/Kconfig
```diff
@@ -172,6 +172,51 @@ config CHIP_FACTORY_RESET_ERASE_NVS
...
+	depends on CHIP_WIFI
+	default 0
+	help
+	  Specifies the delay after a successful connection recovery.
+	  The 0 value means that the connection recovery interval will be cleared
+	  immediately after the connection recovery.
+	  A non-zero value delays the interval reset, which is helpful in the situation
+	  where a link quality to a WiFi access point is poor and device disconnects
```
wiba-nordic, line 217:
```suggestion
	  A non-zero value delays the interval reset, which is helpful when
	  link quality to a Wi-Fi access point is poor and the device disconnects 
```

wiba-nordic (changes requested):

config/nrfconnect/chip-module/Kconfig
markaj-nordic: Right, it's essentially the same. Removed.

wiba-nordic (approved):
Descriptions look good now.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -147,6 +147,10 @@ CHIP_ERROR NrfWiFiDriver::CommitConfiguration()
 
 CHIP_ERROR NrfWiFiDriver::RevertConfiguration()
 {
+    // Abort Connection Recovery if it is in progress during reverting configuration.
+    // This is needed to stop recovery process after failsafe timer expiring.
+    WiFiManager::Instance().AbortConnectionRecovery();
```
Damian-Nordic, line 152: Shouldn't this simply be called by `WiFiManager::Instance().Disconnect()`? IMHO it seems natural that when I disconnect I really mean it, and the connection shouldn't be resurrected on the next timer fire time :)

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -358,9 +360,20 @@ void WiFiManager::ScanDoneHandler(Platform::UniquePtr<uint8_t> data)
...
         }
 
-        // Internal scan is supposed to be followed by connection request
+        // Internal scan is supposed to be followed by a connection request if the SSID has been found
         if (Instance().mInternalScan)
         {
+            if (!Instance().mSsidFound && !Instance().mRecoveryTimerAborted)
+            {
+                ChipLogProgress(DeviceLayer, "No requested SSID found");
+                delete[] rawData;
```
Damian-Nordic, line 369: Hmm. Does it mean we have a leak in the line 389? How about packing `rawData` again in `UniquePtr` to prevent from leaking the data when adding new code here?

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -474,5 +491,53 @@ void WiFiManager::PostConnectivityStatusChange(ConnectivityChange changeType)
...
+{
+    // If kConnectionRecoveryMaxOverallInterval has a non-zero value prevent endless re-scan.
+    if (0 != kConnectionRecoveryMaxRetries && (++Instance().mConnectionRecoveryCounter >= kConnectionRecoveryMaxRetries))
+    {
+        Instance().AbortConnectionRecovery();
+        return;
+    }
+
+    // Prevent executing Scan if it has been aborted by the failsafe timeout occurrence.
+    if (!Instance().mRecoveryTimerAborted)
```
Damian-Nordic, line 521: This must be true, because the timer is cancelled when `mRecoveryTimerAborted` is set, right?

Damian-Nordic (approved):
Feel free to fix these nits in the follow up task :)

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
markaj-nordic: Indeed it makes total sense to call this in the Disconnect() as well, but here we need to avoid further retires when the fail safe timer expires, so it should not be dependent on the connection state (we call Disconnect() here only when the device is connected to the network), or do I miss something? Maybe we can call Disconnect regardless of the state, that would address all use cases. Or move the check to the Disconnect itself...

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: You mean that the Recover will not be called all at all when the `AbortConnectionRecovery` is called hence the check is redundant?

Damian-Nordic: Yes, exactly.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
Damian-Nordic: Ok, so let's leave it for now, but IMHO it would be cleaner if this retry logic was hidden inside WiFiManager so that the caller does not need to know all the dependencies between methods.

markaj-nordic: I agree, this should be the goal. I hope this can be fixed in the next iteration.

markaj-nordic (approved):

PR #207, opened 2023-02-08
[nrf noup] Remove Android Chip Tool build from release workflow

ArekBalysNordic:
Android Chip Tool is no longer maintained so it must be removed from the release workflow.


kkasperczyk-no (approved):
Did you try to run the workflow to check if it passes?

ArekBalysNordic:
> Did you try to run the workflow to check if it passes?

image
Looks OK.

PR #214, opened 2023-02-15
[nrf noup] Remove dependency to openssl from CHIP tool build for linux

ArekBalysNordic:
Currently, the chip-tool that is released as a part of sdk-connectedhomeip has dependency to openssl.
The issue is that Ubuntu 20.04 and Ubuntu 22.04 use different versions of OpenSSL and the version built under 20.04 does not work under 22.04 (without some hacks to install the old openssl version) and vice verse.

- Removed the dependency to the dynamic library (OpenSSL) to build Chip-Tool with the mbedTLS.

### DIFF
--- a/.github/workflows/release_tools.yaml
+++ b/.github/workflows/release_tools.yaml
@@ -63,13 +63,13 @@ jobs:
             - name: Build x64 CHIP Tool with debug logs enabled
               timeout-minutes: 10
               run: |
-                  scripts/run_in_build_env.sh "gn gen out/chiptool_x64_debug --args='chip_mdns=\"platform\" symbol_level=0'"
+                  scripts/run_in_build_env.sh "gn gen out/chiptool_x64_debug --args='chip_mdns=\"platform\" chip_crypto=\"mbedtls\" symbol_level=0'"
                   scripts/run_in_build_env.sh "ninja -C out/chiptool_x64_debug chip-tool"
                   mv out/chiptool_x64_debug/chip-tool out/chiptool_x64_debug/chip-tool-debug
             - name: Build x64 CHIP Tool with debug logs disabled
               timeout-minutes: 10
               run: |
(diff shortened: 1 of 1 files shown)


src/crypto/BUILD.gn
```diff
@@ -25,6 +25,8 @@ if (chip_crypto == "") {
       current_os == "zephyr" || current_os == "mbed" || current_os == "webos" ||
       current_os == "cmsis-rtos") {
     chip_crypto = "mbedtls"
+  } else if (current_os == "linux") {
+    chip_crypto = "mbedtls"
```
LuDuda, line 29: Can we try to set `chip_crypto` during building the release artifacts, and left `openssl` here?

ArekBalysNordic: OK, let's check that :)

ArekBalysNordic: https://github.com/ArekBalysNordic/sdk-connectedhomeip/actions/runs/4193817009/jobs/7271181306

ArekBalysNordic:
Provided changes also to OTA-Provider build. Checking release workflow:
https://github.com/ArekBalysNordic/sdk-connectedhomeip/actions/runs/4195307001

src/tools/chip-cert/BUILD.gn
```diff
@@ -37,7 +37,9 @@ executable("chip-cert") {
     "chip-cert.h",
   ]
 
-  public_configs = [ "${chip_root}/src/crypto:openssl_config" ]
+  if (chip_crypto == "openssl") {
```
LuDuda, line 40: Do we need this change at all?

LuDuda (approved):
Thanks :+1:

src/tools/chip-cert/BUILD.gn
ArekBalysNordic: Right, we don't build chip-cert in the release workflow, so we don't need it.

PR #217, opened 2023-02-16
[nrf noup] Mapped the WiFi security type from Zephyr to Matter

ArekBalysNordic:
The Security type attribute was wrongly cast in the WiFi Manager. Used switch-case-based mapping between two enum types (Zephyr-Matter) to make sure that we properly cast between two different types.

### DIFF
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -99,6 +99,27 @@ uint8_t MapToMatterWiFiVersionCode(wifi_link_mode wifiVersion)
     return (static_cast<uint8_t>(wifiVersion) - 1);
 }
 
+// Matter expectations towards Wi-Fi security type codes are unaligned with
+// what wpa_supplicant provides. This function maps supplicant codes
+// to the ones defined in the Matter spec (11.14.3.1. SecurityType enum)
+uint8_t MapToMatterSecurityType(wifi_security_type securityType)
+{
+    switch (securityType)
+    {
+    case WIFI_SECURITY_TYPE_NONE:
+        return EMBER_ZCL_SECURITY_TYPE_NONE;
+    case WIFI_SECURITY_TYPE_PSK:
+    case WIFI_SECURITY_TYPE_PSK_SHA256:
+        return EMBER_ZCL_SECURITY_TYPE_WPA2;
+    case WIFI_SECURITY_TYPE_SAE:
+        return EMBER_ZCL_SECURITY_TYPE_WPA3;
+    default:
+        break;
+    }
+
+    return EMBER_ZCL_SECURITY_TYPE_UNSPECIFIED;
+}
+
 } // namespace
 
 const Map<wifi_iface_state, WiFiManager::StationStatus, 10>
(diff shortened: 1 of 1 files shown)


src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -99,6 +99,34 @@ uint8_t MapToMatterWiFiVersionCode(wifi_link_mode wifiVersion)
     return (static_cast<uint8_t>(wifiVersion) - 1);
 }
 
+// Matter expectations towards Wi-Fi security type codes are unaligned with
+// what wpa_supplicant provides. This function maps supplicant codes
+// to the ones defined in the Matter spec (11.14.3.2. SecurityType enum)
+uint8_t MapToMatterSecurityType(wifi_security_type securityType)
+{
+    if (securityType < WIFI_SECURITY_TYPE_NONE || securityType > WIFI_SECURITY_TYPE_SAE ||
```
kkasperczyk-no, line 107: How about the following change to avoid problems in case of any changes in this enum in the future?
```suggestion
    if (securityType < WIFI_SECURITY_TYPE_NONE || securityType > WIFI_SECURITY_TYPE_MAX ||
```

kkasperczyk-no (approved):

src/platform/nrfconnect/wifi/WiFiManager.cpp
Damian-Nordic: Move this to `default:` instead of relying on the order of the enums?

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -99,6 +99,34 @@ uint8_t MapToMatterWiFiVersionCode(wifi_link_mode wifiVersion)
...
+    case WIFI_SECURITY_TYPE_PSK:
+    case WIFI_SECURITY_TYPE_PSK_SHA256:
+        return EMBER_ZCL_SECURITY_TYPE_WPA2;
+    case WIFI_SECURITY_TYPE_SAE:
+        return EMBER_ZCL_SECURITY_TYPE_WPA3;
+    default:
+        break;
+    }
+
+    return (static_cast<uint8_t>(securityType) - 1);
```
Damian-Nordic, line 127: What is this supposed to do?

ArekBalysNordic: It's not needed here, I will replace it with EMBER_ZCL_SECURITY_TYPE_UNSPECIFIED.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -99,6 +99,34 @@ uint8_t MapToMatterWiFiVersionCode(wifi_link_mode wifiVersion)
     return (static_cast<uint8_t>(wifiVersion) - 1);
 }
 
+// Matter expectations towards Wi-Fi security type codes are unaligned with
+// what wpa_supplicant provides. This function maps supplicant codes
+// to the ones defined in the Matter spec (11.14.3.2. SecurityType enum)
```
markaj-nordic, line 104: The SecurityType enum is placed in the 11.14.3.1

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: This is the default mapping for the WiFi version enumeration based on the spec. I'm not sure if this applies to the security mode as well.

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Maybe we could extend the Map implementation with some rage checks (e.g. add a `bool contains()` method) and use his utility for all required mappings. But this is just a nit.

Damian-Nordic (approved):

markaj-nordic (approved):

PR #224, opened 2023-02-27
[nrf noup] Support disabling Wi-Fi LPM during the OTA DFU

markaj-nordic:
Added subclass of OTAImageProcessorImpl whcih implements Wi-Fi specific optimization for LPM which should be disabled during OTA due to the high throughput requirements. This allows applications to use the OTAImageProcessorImplWiFi to do the job.

### DIFF
--- a/src/platform/nrfconnect/OTAImageProcessorImplWiFi.h
+++ b/src/platform/nrfconnect/OTAImageProcessorImplWiFi.h
@@ -1,0 +1,63 @@
+/*
+ *    Copyright (c) 2023 Project CHIP Authors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#pragma once
+
+#include "OTAImageProcessorImpl.h"
+#include "wifi/WiFiManager.h"
+
+namespace chip {
+
+namespace DeviceLayer {
+
+class OTAImageProcessorImplWiFi : public OTAImageProcessorImpl
+{
+public:
+    explicit OTAImageProcessorImplWiFi(ExternalFlashManager * flashHandler = nullptr) : OTAImageProcessorImpl(flashHandler) {}
+
+    CHIP_ERROR PrepareDownload() override
+    {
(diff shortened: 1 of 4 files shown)


src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -564,5 +564,34 @@ void WiFiManager::AbortConnectionRecovery()
...
+    if (net_mgmt(NET_REQUEST_WIFI_PS_CONFIG, iface, &currentConfig, sizeof(currentConfig)))
+    {
+        ChipLogError(DeviceLayer, "Get current low power mode config request failed");
+        return CHIP_ERROR_INTERNAL;
+    }
+
+    if ((currentConfig.enabled == WIFI_PS_ENABLED && onoff == false) ||
+        (currentConfig.enabled == WIFI_PS_DISABLED && onoff == true))
+    {
+        wifi_ps_params params{ .enabled = onoff ? WIFI_PS_ENABLED : WIFI_PS_DISABLED };
```
Damian-Nordic, line 582: Would it make sense to modify `currentConfig` instead of creating a new structure, to preserve other PS settings? Like whether it's WMM or legacy.

markaj-nordic: Not sure if I get this right but if you mean reusing the currentConfig in the `net_mgmt(NET_REQUEST_WIFI_PS,...)` call it's a different API: `wifi_ps_config` vs. `wifi_ps_params`. But we can preserve the current power save mode configuration for the future purposes, i.e. by by extending the ConnectionParams structure with `wifi_ps_config` field.

Damian-Nordic: ah, right. I thought it's a single structure used for both APIs. Nevermind then.

Damian-Nordic (approved):

LuDuda (approved):

src/platform/nrfconnect/OTAImageProcessorImplWiFi.h
```diff
@@ -0,0 +1,63 @@
...
+namespace DeviceLayer {
+
+class OTAImageProcessorImplWiFi : public OTAImageProcessorImpl
+{
+public:
+    explicit OTAImageProcessorImplWiFi(ExternalFlashManager * flashHandler = nullptr) : OTAImageProcessorImpl(flashHandler) {}
+
+    CHIP_ERROR PrepareDownload() override
+    {
+        CHIP_ERROR err = WiFiManager::Instance().SetLowPowerMode(false);
```
Damian-Nordic, line 33: The approach is correct though we may think in the future about incorporating events like https://github.com/project-chip/connectedhomeip/commit/7e728bc1bd0275f1dff8d75b2214c745047f6e01 (once that is merged to sdk-connectedhomeip) to do certain actions on OTA state changes to make the solution more extensible.

markaj-nordic: Thanks for pointing this out @Damian-Nordic

PR #229, opened 2023-03-02
[nrf noup] Provide configs for Matter NUS feature

ArekBalysNordic:
Matter NUS feature allows to add of BLE Nordic UART Service and register commands which allow controlling the Matter device via BLE.

This can be useful when the device loses connection with a Matter Controller but it should be controlled in another way for e.g. a Door Lock.

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig.features
+++ b/config/nrfconnect/chip-module/Kconfig.features
@@ -73,6 +73,58 @@ config SPI_NOR_FLASH_LAYOUT_PAGE_SIZE
 
 endif # CHIP_SPI_NOR
 
+config CHIP_NUS
+	bool "Enable Nordic UART service for Matter purposes"
+	select BT_NUS
+	select BT_SMP
+	select BT_NUS_AUTHEN
+	select BT_SETTINGS
+	help
+	  Enables Nordic UART service (NUS) for Matter samples.
+	  Using NUS service you can control a Matter sample using pre-defined BLE commands 
+	  and do defined operations. The CHIP NUS service can be useful to keep communication
+	  with a smart home device when a connection within Matter network is lost.
+
+if CHIP_NUS
+
+config BT_RX_STACK_SIZE
+	default 1536
+
+config SYSTEM_WORKQUEUE_STACK_SIZE
+	default 2048
+
+config BT_MAX_CONN
+	default 2
+
+config BT_MAX_PAIRED
+	default 2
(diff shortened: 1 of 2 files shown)


config/nrfconnect/chip-module/Kconfig.features
```diff
@@ -73,6 +73,45 @@ config SPI_NOR_FLASH_LAYOUT_PAGE_SIZE
 
 endif # CHIP_SPI_NOR
 
+config CHIP_NUS
```
LuDuda, line 76: We can discuss it, but I don't think that NUS should be combined with Mater in any way apart of application itself. What do you think to move this new config just to Door Lock sample's KConfig.

I would suggest to move

kkasperczyk-no: Moving it to the door lock sample would mean that it will not be used in any other sample. It looks like a quite generic functionality that users could benefit from in any of the samples, right?

markaj-nordic: I guess what Lukasz is trying to convey is that it's not really a Matter feature. I guess Ideally it should be put somewhere which is generic for all samples (in NCS) but not embraced in Matter specific configs.

Damian-Nordic: Maybe the DFU over SMP Kconfig (and the NUS Kconfig) should be put next to the code for that, so somewhere in `examples/platform/nrfconnect` (in the upstream) and in `samples/matter/common` (in NCS). It is a bit confusing that Kconfig.features now contains Matter features like the Wi-Fi connection recovery and features that don't have anything to do with the Matter implementation. Also, as a general rule, I think we should provide an example to our customers that they can do a lot of modifications to Matter applications without touching the Matter SDK (and asking for our assistance).

kkasperczyk-no: We just had a discussion on that during daily meeting. I believe that we started to maintain Kconfig.features in the Matter repo and not in NCS, because we were spending much more time on working with upstream. This way it was much faster to modify Matter code and have the change in place, instead of modifying something in NCS and bumping NCS revision in Matter. Now once our focus moved to NCS, maybe we should just move this Kconfig,features to NCS samples/matter/common and source this file in upstream examples.

LuDuda (approved):

markaj-nordic (approved):

PR #275, opened 2023-04-27
[nrf noup] Introduce workaround to Wi-Fi connection recovery

kkasperczyk-no:
It was observed that sometimes WPA supplicant may stuck in the reconnection process. This change introduces a workaround and fixes for existing recovery mechanism:

* Changed is-Recovery-Aborted approach to is-Recovery-started.
* Distinguished disconnects requested by the application from the disconnects coming from the supplicant.
* Added scheduling recovery procedure if supplicant was not able to re-connect to the network for given amount of time (60 s).

### DIFF
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -205,6 +205,7 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
     mCachedWiFiState    = mWiFiState;
     mWiFiState          = WIFI_STATE_SCANNING;
     mSsidFound          = false;
+    mRecoveryArmed    = true;
 
     if (net_mgmt(NET_REQUEST_WIFI_SCAN, iface, NULL, 0))
     {
@@ -266,6 +267,7 @@ CHIP_ERROR WiFiManager::Disconnect()
     }
     else
     {
+        mDisconnectRequested = true;
         ChipLogDetail(DeviceLayer, "Disconnect requested");
     }
 
@@ -390,15 +392,21 @@ void WiFiManager::ScanDoneHandler(Platform::UniquePtr<uint8_t> data)
         // Internal scan is supposed to be followed by a connection request if the SSID has been found
         if (Instance().mInternalScan)
         {
-            if (!Instance().mSsidFound && !Instance().mRecoveryTimerAborted)
+            if (Instance().mRecoveryArmed)
             {
-                ChipLogProgress(DeviceLayer, "No requested SSID found");
-                auto currentTimeout = Instance().CalculateNextRecoveryTime();
-                ChipLogProgress(DeviceLayer, "Starting connection recover: re-scanning... (next attempt in %d ms)",
-                                currentTimeout.count());
-                Instance().mRecoveryTimerAborted = false;
-                DeviceLayer::SystemLayer().StartTimer(currentTimeout, Recover, nullptr);
-                return;
+                if (!Instance().mSsidFound)
+                {
+                    ChipLogProgress(DeviceLayer, "No requested SSID found");
+                    auto currentTimeout = Instance().CalculateNextRecoveryTime();
+                    ChipLogProgress(DeviceLayer, "Starting connection recover: re-scanning... (next attempt in %d ms)",
+                                    currentTimeout.count());
+                    DeviceLayer::SystemLayer().StartTimer(currentTimeout, Recover, nullptr);
+                    return;
+                }
+                else
+                {
+                    Instance().AbortConnectionRecovery();
+                }
             }
 
             Instance().mWiFiState = WIFI_STATE_ASSOCIATING;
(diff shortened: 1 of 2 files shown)


src/platform/nrfconnect/wifi/WiFiManager.h
```diff
@@ -173,6 +173,7 @@ class WiFiManager
     static constexpr uint32_t kConnectionRecoveryMaxIntervalMs     = CONFIG_CHIP_WIFI_CONNECTION_RECOVERY_MAXIMUM_INTERVAL;
     static constexpr uint32_t kConnectionRecoveryJitterMs          = CONFIG_CHIP_WIFI_CONNECTION_RECOVERY_JITTER;
     static constexpr uint32_t kConnectionRecoveryMaxRetries        = CONFIG_CHIP_WIFI_CONNECTION_RECOVERY_MAX_RETRIES_NUMBER;
+    static constexpr uint32_t kSupplicantReconnectionTimeoutMs     = 60000;
```
ArekBalysNordic, line 176: It is defined somewhere, measured, or just set regarding observations?

kkasperczyk-no: It's just something we need to decide. I don't think there are any measurements we could do. This is the value that defines how long time are we ok to give supplicant to fix our connection. After this time we will disconnect and start re-scanning based on our recovery mechanism.

kkasperczyk-no: If you or @markaj-nordic have some opinion regarding this value we can change it. 60 s in my opinion was big enough to give module a time to reconnect during usual network hiccups and not too big, so user would be annoyed that the device is not available for too long in case supplicant would stuck

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -205,6 +205,7 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
     mCachedWiFiState    = mWiFiState;
     mWiFiState          = WIFI_STATE_SCANNING;
     mSsidFound          = false;
+    mRecoveryStarted    = true;
```
markaj-nordic, line 208: Seems weird that we start recovery in the scan request.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -504,12 +512,36 @@ void WiFiManager::ConnectHandler(Platform::UniquePtr<uint8_t> data)
...
+    }
+    else
+    {
+        // Workaround: schedule the application level connection recovery in kSupplicantReconnectionTimeoutMs to give WPA supplicant
+        // some time to restore it.
+        if (!Instance().mRecoveryStarted)
+        {
+            Instance().mRecoveryStarted = true;
+            DeviceLayer::SystemLayer().StartTimer(
+                System::Clock::Milliseconds32(kSupplicantReconnectionTimeoutMs),
```
markaj-nordic, line 532: I think we can save some time if we cancel this timer in the `WiFiManager::ConnectHandler` right away (now we need to wait for another `Recover()` call to issue `AbortConnectionRecovery()`).

src/platform/nrfconnect/wifi/WiFiManager.h
markaj-nordic: Yehh, I don't have a strong opinion on that, maybe from the user perspective 60s is better then 120 :)

src/platform/nrfconnect/wifi/WiFiManager.cpp
kkasperczyk-no: Why? I mean, maybe the field name is not the best, as it's not like we start recovery here, but rather enable possibility of starting it. Would naming it `mRecoveryEnabled` be better for you?

src/platform/nrfconnect/wifi/WiFiManager.cpp
kkasperczyk-no: No, it will be aborted in the `ScanDoneHandler` after checking if SSID was found: https://github.com/nrfconnect/sdk-connectedhomeip/pull/275/files#diff-360bd1fbca9712db3463cc36c37be05596a04f08eab2f947cee8bf1c07ad6835R408

LuDuda (approved):
Thanks  👍

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Sounds better. Or, following Matter FailSafe fashion - `mRecoveryArmed` :). It's a nit, we can leave it as it is unless you are introducing more changes anyway.

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Yes but... the original recovery mechanism was introduced to have a chance to find the SSID which is not be present immediately after reboot for whatever reason (i.e. it does not care about the network credentials). Here, we are handling the connection instabilities at *runtime*, so I would expect the recovery to be finalized only when the connection is back again instead of relying solely on the SSID being found.

src/platform/nrfconnect/wifi/WiFiManager.cpp
kkasperczyk-no: > Or, following Matter FailSafe fashion - mRecoveryArmed :).

I like it! Done.

src/platform/nrfconnect/wifi/WiFiManager.cpp
kkasperczyk-no: Discussed F2F, we will create a follow-up task for that, which could be connected with the WiFiManager refactor.

markaj-nordic (approved):

PR #281, opened 2023-05-12
Tag 1.1.0.1 followup

ArekBalysNordic:
Provided all needed changes before NCS 2.4.0 RC1 TAG:

- [x] Repaired CI unit tests
- [x] CP from upstream: Fix handling of certificate validity policy in CHIPDeviceControllerFactory.

### DIFF
--- a/src/protocols/BUILD.gn
+++ b/src/protocols/BUILD.gn
@@ -13,9 +13,14 @@
 # limitations under the License.
 
 import("//build_overrides/chip.gni")
+import("${chip_root}/src/platform/device.gni")
 
 static_library("protocols") {
   output_name = "libChipProtocols"
+  
+  if(chip_device_platform == "linux") {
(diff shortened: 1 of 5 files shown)


src/protocols/BUILD.gn
```diff
@@ -17,6 +17,8 @@ import("//build_overrides/chip.gni")
 static_library("protocols") {
   output_name = "libChipProtocols"
 
+  defines = ["CHIP_DEVICE_CONFIG_ENABLE_COMMISSIONER_DISCOVERY_CLIENT=1"]
```
kkasperczyk-no, line 20: Why do we need it? If it's needed to fix unit tests maybe we should enable it in conditional way. Now it will be also enabled for our platform and I don't think we want it.

ArekBalysNordic: We need it due to changes regarding removing POSIX api, but right, it should be enabled in a conditional way. I will add it.

kkasperczyk-no (approved):

PR #286, opened 2023-05-22
[nrf noup] Add a workaround to WiFi recovery

ArekBalysNordic:
During the recovery procedure, if the net_mgmt function returns the value "-16" the device cannot recover the WiFi network until the reboot is called.

This workaround allows the device to try to rescan the WiFi network even if the net_mgmt function returned an error. This workaround works only, when the WiFi recovery mechanism is active.


src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -208,9 +208,23 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
...
 
-    if (net_mgmt(NET_REQUEST_WIFI_SCAN, iface, NULL, 0))
+    int ret = net_mgmt(NET_REQUEST_WIFI_SCAN, iface, NULL, 0);
+
+    if (ret)
     {
-        ChipLogError(DeviceLayer, "Scan request failed");
+        ChipLogError(DeviceLayer, "Scan request failed %d", ret);
+        if(mRecoveryActive){
+            /* Wi-Fi driver returned an error during recovery.
```
kkasperczyk-no, line 217: I don't like this workaround for several reasons:
1. In theory the current state logic should be correct, so it would be better to investigate the issue in the supplicant code, instead of creating a workaround of solution that is already a workaround. Nevertheless I know we have to do something to fix this and we don't have much time, so let's say I'm ok with that.
2. We should at least try to analyze the returned error code to decide, whether we want to retry or not. Now we may fall into a loop of pointless retries, e.g. because the error code informs about something that will not be fixed just by waiting some time.
3. I don't think that the approach with incrementing the re-try time is a good idea for this particular case. Even if something is messed up with the supplicant state and we want to give it some more time to clean it, I think we should try to ask once again, but not more. If it wasn't fixed in a first few ms, I don't think it will be fixed ever. Maybe in case of getting error on a retry we should just crash and reboot the app to make sure we will not stuck in invalid networking state. Any thoughts @markaj-nordic?

kkasperczyk-no (approved):

src/platform/nrfconnect/wifi/WiFiManager.h
```diff
@@ -246,6 +246,8 @@ class WiFiManager
     uint32_t mConnectionRecoveryTimeMs{ kConnectionRecoveryMinIntervalMs };
     bool mRecoveryArmed{ false };
     bool mApplicationDisconnectRequested{ false };
+    // TODO Workaround for recovery mechanism to wait before the next scan request until the WiFi supplicant is not busy.
+    System::Clock::Seconds32 WifiSupplicantWorkaroundTime {8};
```
markaj-nordic, line 250: Please keep the notation scheme, add make it const or constexpr. Looks like the formatting is also a bit off
```suggestion
    static constexpr System::Clock::Seconds32 kWifiSupplicantWorkaroundTime{ 8 };
```

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -207,10 +207,29 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
     mWiFiState          = WIFI_STATE_SCANNING;
     mSsidFound          = false;
     mRecoveryArmed      = true;
+    // TODO Workaround for recovery mechanism to wait before the next scan request until the WiFi supplicant is not busy.
+    static bool workaroundDone;
```
markaj-nordic, line 211: I know it's already `false` but maybe better initialize it explicitly anyway.

ArekBalysNordic: Static is always initialized with 0 (false), isn't it? Or I can set it to false just before calling reboot if something goes wrong.

markaj-nordic (approved):

ArekBalysNordic:
There were some linker problems, repaired, please look at it one more time.

PR #294, opened 2023-06-09
Move default security configs to nrf 

SebastianBoe:
Move default security configs to nrf to ensure they sourced before
nrf_security configs also after nrf_security has been moved from
nrfxlib to nrf.


SebastianBoe:
No longer necessary.

PR #295, opened 2023-06-12
[nrf noup] Fix ram_report and rom_report

LuDuda:
Stop filtering out gdwarf-4 when passing flags from Zephyr to Matter's GN build system. Older pyelftools versions are not able to parse DWARF5 format, so ram and rom report would not be generated.

Re-applying #165 on a new codebase.

### DIFF
--- a/config/common/cmake/make_gn_args.py
+++ b/config/common/cmake/make_gn_args.py
@@ -28,7 +28,11 @@ GN_CFLAG_EXCLUDES = [
     '-fno-reorder-functions',
     '-ffunction-sections',
     '-fdata-sections',
-    '-g*',
+    '-g',
+    '-g0',
+    '-g1',
+    '-g2',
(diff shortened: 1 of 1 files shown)


markaj-nordic (approved):
Was this change reverted in the past? Just wondering why we need to reapply it.

Damian-Nordic (approved):

LuDuda:
> Was this change reverted in the past? Just wondering why we need to reapply it.

Yes, the file on which we had fix was moved out of nrfconnect/zephyr directory to common folder in upstream. During up merge to Matter SDK 1.1.0.1 it was missed.

ArekBalysNordic:
As a result of the recent conversation, I've added this commit there: https://github.com/nrfconnect/sdk-connectedhomeip/pull/296. Can we close this PR?

PR #297, opened 2023-06-13
All fixes needed for NCS 2.4.1 release.

ArekBalysNordic:
This is cumulative PR to collect all needed bug fixes for NCS 2.4.1 release.

- [x] Cherry-Pick: https://github.com/project-chip/connectedhomeip/commit/8db0d655a074c8b3d5539c7847fc54f8448cb5a3 
- [x] Cherry-Pick: https://github.com/project-chip/connectedhomeip/commit/0f8fac1f86a91fd3fe4e1056d546ef00de72e46f
- [x] Cherry-Pick: https://github.com/nrfconnect/sdk-connectedhomeip/pull/295/commits/39c99f79e34f2b0ebfaeb70e9800d45a2abf14be
- [ ] Cherry-Pick: https://github.com/nrfconnect/sdk-connectedhomeip/commit/21ee935b713bd94e780315a3f9366b2da350422f
- [ ] Optionally fixes for Sheliak
- [ ] Known issues to v2.4.0 are added for the important bugs (like clusters configuration, wrong links, etc).  
- [ ]


src/platform/nrfconnect/DiagnosticDataProviderImplNrf.cpp
```diff
@@ -41,6 +41,24 @@ DiagnosticDataProviderImplNrf & DiagnosticDataProviderImplNrf::GetDefaultInstanc
     return sInstance;
 }
 
+
+// Disable all Watermarks methods since the feature map is set to 0
+CHIP_ERROR DiagnosticDataProviderImplNrf::ResetWatermarks()
```
Damian-Nordic, line 46: What does it mean that the feature map is set to 0? It is application-specific so a user might want to use this feature.

doublemis1: So for our samples we don't want to support reset watermarks, since there always are issues with that (do we properly handle the reset watermarks and watermarks heap?). Hence we want to disable feature watermarks in Software diagnostics cluster. However due to implementation of software diagnostics server in matter sdk, we cannot just disable all watermarks command/attributes and set feature-map to 0 on zap level.
https://github.com/project-chip/connectedhomeip/issues/27099

Damian-Nordic (approved):
I think that it would be better to keep the support for high watermarks, but e.g. add a config to platform/Zephyr/CHIPDevicePlatformConfig.h to disable it by default. But I don't know if any customers actually use it.

ArekBalysNordic:
> I think that it would be better to keep the support for high watermarks, but e.g. add a config to platform/Zephyr/CHIPDevicePlatformConfig.h to disable it by default. But I don't know if any customers actually use it.

Ok, that sounds good :)

kkasperczyk-no (approved):

PR #301, opened 2023-06-23
config: nrfconnect: Remove PSA_WANT default n setting

joerchan:
With PSA_WANT configurations being default n then these default values are no longer needed for chip-module.

NCSDK-18031

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig.defaults
+++ b/config/nrfconnect/chip-module/Kconfig.defaults
@@ -355,39 +355,6 @@ config MBEDTLS_GCM_C
 config MBEDTLS_RSA_C
     default n
 
-config PSA_WANT_KEY_TYPE_CHACHA20
-    default n
-
-config PSA_WANT_ALG_GCM
-    default n
-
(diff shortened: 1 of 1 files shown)


config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -355,39 +355,6 @@ config MBEDTLS_GCM_C
 config MBEDTLS_RSA_C
     default n
 
-config PSA_WANT_KEY_TYPE_CHACHA20
```
Damian-Nordic, line 358: Does it mean we now should explicitly enable all crypto algorithms used by Matter?

Damian-Nordic (approved):

config/nrfconnect/chip-module/Kconfig.defaults
joerchan: That is the goal. But for now only on the PSA configurations.
The mbedtls ones are not touched in this.

SebastianBoe (approved):

ArekBalysNordic (approved):

PR #303, opened 2023-07-10
[nrf fromlist] bypass support for IPv4

markaj-nordic:
This change allows to build nrfconnect SDK samples with Matter over IPv6 but IPv4 still available for third party development within the same application.


config/nrfconnect/chip-module/Kconfig
```diff
@@ -259,3 +259,11 @@ config CHIP_LOG_SIZE_OPTIMIZATION
 	  information that is too detailed to be used in most cases. You can find
 	  full configuration enabled by this option in the
 	  platform/nrfconnect/CHIPPlatformConfig.h file.
+
+config CHIP_BYPASS_IPV4
```
Damian-Nordic, line 263: Maybe rename it to `CHIP_IPV4` and use `default n` (+ `depends on NET_IPV4`). Then you could directly set `chip_inet_config_enable_ipv4` to `CONFIG_CHIP_IPV4`.

markaj-nordic: Ok, thanks for the hint, now indeed the `chip_inet_config_enable_ipv4` is defaulted to TRUE, so needed an explicit FALSE in CMakeListst.txt.

Damian-Nordic (approved):

kkasperczyk-no (approved):

PR #311, opened 2023-07-31
Switch to PSA crypto

ArekBalysNordic:
Draft PR with PSA Crypto enabled within Matter and OpenThread.

- Selected OpenThread security PSA Crypto background
- Switched to PSAOperationalKeystore when CHIP_CRYPTO_PSA is enabled
- Changed definitions from CONFIG_NORDIC_SECURITY_BACKEND to
CONFIG_NRF_SECURITY to avoid using MBEDTLS Legacy.

WARNING: PSA Crypto is selected to y by default!


config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -288,40 +289,38 @@ config MBEDTLS_ENABLE_HEAP
...
-    default 15360
-
-config MBEDTLS_TLS_LIBRARY
-    default y
+    default 15360 if !CHIP_CRYPTO_PSA
+    default 32768 if CHIP_CRYPTO_PSA
 
 config NRF_SECURITY_ADVANCED
-    default y
+    default y if !CHIP_CRYPTO_PSA
```
LuDuda, line 296: nit: would it work to have `if !CHIP_CRYPTO_PSA` in line 294 for all mbedtls related config?

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -288,40 +289,38 @@ config MBEDTLS_ENABLE_HEAP
     default y
 
 config MBEDTLS_HEAP_SIZE
-    default 15360
-
-config MBEDTLS_TLS_LIBRARY
-    default y
+    default 15360 if !CHIP_CRYPTO_PSA
+    default 32768 if CHIP_CRYPTO_PSA
```
Damian-Nordic, line 293: Have we reported this growth to nrf_security team? I think they should be aware of that and try to optimize that?

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -288,40 +289,38 @@ config MBEDTLS_ENABLE_HEAP
...
-    default y
+    default y if !CHIP_CRYPTO_PSA
 
 config MBEDTLS_AES_C
-    default y
+    default y if !CHIP_CRYPTO_PSA
 
 config MBEDTLS_ECP_C
-    default y
+    default y if !CHIP_CRYPTO_PSA
```
Damian-Nordic, line 302: Don't we need to enable any PSA_WANT_ALG_XXX here? We may rely on OT here, but it would be good to have them enabled directly in Matter as well to support Wi-Fi and possibly other platforms in the future as well.

Damian-Nordic (approved):
Have you maybe checked what is the flash increase after switching to PSA in the current form?

config/nrfconnect/chip-module/Kconfig.defaults
ArekBalysNordic: I need to perform more tests. For now, I realized that 16 kB of Heap is too low to run PSA Crypto, 32kB is the secure value. Probably, it can be reduced right now, but first I would like to check the actual demand.

ArekBalysNordic:
> Have you maybe checked what is the flash increase after switching to PSA in the current form?

@Damian-Nordic 
With Joiner disabled and PSA Crypto enabled we have 20 kB less Flash occupancy in comparison to MbedTLS Legacy. But for now, RAM usage increased ~14 kB  - This is to be verified regarding MbedTLS HEAP size.

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -267,7 +267,8 @@ endif # CHIP_WIFI
 # ==============================================================================
 
 choice OPENTHREAD_SECURITY
-    default OPENTHREAD_NRF_SECURITY_CHOICE if NET_L2_OPENTHREAD
+    default OPENTHREAD_NRF_SECURITY_CHOICE if !CHIP_CRYPTO_PSA
```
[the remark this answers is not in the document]
ArekBalysNordic: Ok, good to know, thanks!

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -284,18 +285,56 @@ config MBEDTLS_ENABLE_HEAP
     default y
 
 config MBEDTLS_HEAP_SIZE
-    default 8192
+    default 15360 if !CHIP_CRYPTO_PSA
```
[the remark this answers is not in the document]
ArekBalysNordic: Oh, this is a mistake

Damian-Nordic (approved):

config/zephyr/Kconfig
```diff
@@ -267,6 +267,7 @@ config CHIP_OPERATIONAL_TIME_SAVE_INTERVAL
 
 config CHIP_CRYPTO_PSA
 	bool "Use PSA crypto API for cryptographic operations"
+	select EXPERIMENTAL
```
kkasperczyk-no, line 270: Remember it's a common Zephyr module. I'm not sure we should select EXPERIMENTAL for all platforms. Maybe we should create a dummy nRF only kconfig that depends on CHIP_CRYPTO_PSA and selects EXPERIMENTAL.

ArekBalysNordic: Oh, right. But we have assigned it as experimental in another config related to PSA, so it can be removed from there.

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -285,17 +284,69 @@ config MBEDTLS_ENABLE_HEAP
...
 
-config MBEDTLS_ECP_C
+config PSA_WANT_KEY_TYPE_ECC_KEY_PAIR
     default y
 
-config MBEDTLS_ECP_DP_SECP256R1_ENABLED
+config MBEDTLS_PSA_CRYPTO_STORAGE_C
+    default y if !BUILD_WITH_TFM
+
+config PSA_WANT_GENERATE_RANDOM
```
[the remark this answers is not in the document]
ArekBalysNordic: I'm going to remove most of those kconfigs since they are already selected by the `OPENTHREAD_NRF_SECURITY_PSA` kconfig. The others are strictly related to nrfconnect.

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -285,17 +284,40 @@ config MBEDTLS_ENABLE_HEAP
 
 config MBEDTLS_HEAP_SIZE
     default 8192
+    
+# Enable PSA Crypto dependencies for Matter
 
-config NRF_SECURITY_ADVANCED
+config CHIP_CRYPTO_PSA
+    default n
+    select EXPERIMENTAL
```
Damian-Nordic, line 292: Add this to the original CHIP_CRYPTO_PSA definition and remove this one?

Damian-Nordic (approved):

config/nrfconnect/chip-module/Kconfig.defaults
kkasperczyk-no, line 292: Sorry for bothering about this one :) But I think we cannot do it like this. Setting it here means that customer that doesn't use our defaults in the application (which is quite possible, because we source them only in our samples) will not get this warning. Can't we just add some dummy Kconfig without prompt, like:

```
config CHIP_CRYPTO_PSA_NRF_EXPERIMENTAL
     select EXPERIMENTAL
     depends on CHIP_CRYPTO_PSA
```
?

ArekBalysNordic: The original Kconfig is defined in Zephyr's config space, so it will impact Telink also if we go upstream with it. Do you think it is ok?

kkasperczyk-no: I requested removing it from there :) It's because I thought it's not ok to select EXPERIMENTAL for all Zephyr platforms, while only we (nrf platform) consider it experimental. I suggested something different here: https://github.com/nrfconnect/sdk-connectedhomeip/pull/311#discussion_r1432565531

ArekBalysNordic: Actually, technically, we can, and we overwrite CHIP_CRYPTO_PSA which is defined in Zephyr's config space. The other thing is that is clear and OK :)

Damian-Nordic: Oh, I see, let's leave it as it is, but `default n` is redundant, so it can be removed at least :)

ArekBalysNordic: Eventually, I can remove it at all, and base on the OPENTHREAD_CRYPT_PSA kconfig, that is also experimental :)

kkasperczyk-no: But maybe I care about the others too much and we should just set it in the Zephyr, as it was.

kkasperczyk-no: > Actually, technically, we can, and we overwrite CHIP_CRYPTO_PSA which is defined in Zephyr's config space. The other thing is that is clear and OK :)

Technically we can do everything, but the point is that it should print a warning to customer and it will not if defaults will not be loaded. I like the Damian's idea. We can override it in nrf space.

ArekBalysNordic: Ok, discussed with the Team, that we're removing experimental at all.

ArekBalysNordic:
@kkasperczyk-no @Damian-Nordic I've removed the experimental from the CHIP_CRYPTO_PSA kconfig. Please look at the newest version of the commit once again.

kkasperczyk-no (approved):

PR #312, opened 2023-08-01
[nrf noup] Repair wrong casting for CSL period setting

ArekBalysNordic:
After the API change, there should not be cast to uint16_t in the otLinkSetCslPeriod function anymore.


markaj-nordic (commented):
@ArekBalysNordic can you briefly describe why such a cast was needed and why it's not needed now?

ArekBalysNordic:
@markaj-nordic API for the otLinkSetCslPeriod has been changed recently. Before the second argument was set as uint16_t because the CSL interval was counted regarding symbols, but now we need to provide a value in us, so the API accepts now uint32_t. The casting there was added due to warnings in the compilation because we wanted to provide a uint32_t value to the uint16_t argument, and now it is not needed anymore since the interval is aligned to uint32_t.

canisLupus1313 (approved):

kkasperczyk-no (approved):

markaj-nordic (approved):

PR #314, opened 2023-08-08
[nrf noup] Enable Factory Data by default for all samples in NCS

ArekBalysNordic:
This PR enables:
- Support for Factory Data for each build.
- Merging the generated factory data set with firmware.
- Generation onboarding codes to the build directory.

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig
+++ b/config/nrfconnect/chip-module/Kconfig
@@ -127,7 +127,6 @@ config CHIP_FACTORY_DATA_CUSTOM_BACKEND
 
 config CHIP_FACTORY_DATA_BUILD
 	bool "Generate factory data contents"
-	default n
 	help
 	  Enables generation of the factory data contents during the building. This
 	  option requires the factory_data partition in Partition Manager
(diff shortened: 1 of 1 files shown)


config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -167,6 +167,17 @@ config BT_DEVICE_NAME_GATT_WRITABLE
 config CHIP_NFC_COMMISSIONING
     default y
 
+# Enable Factory Data support by default
+
+config CHIP_FACTORY_DATA
```
doublemis1, line 172: automatic factory data merge is enabled too?

ArekBalysNordic: yes, it is enabled by default when CHIP_FACTORY_DATA_BUILD is set to y.

Damian-Nordic, line 172: It may cause trouble to customers who include `Kconfig.defaults` but do not have the factory data partition. Not sure if you considered that and accepted the risk, but just saying :)

LuDuda (approved):

PR #315, opened 2023-08-10
[nrf noup] Align finding Python3 executable to NCS requirements.

ArekBalysNordic:
We should try to find Python3 instead of Python within Cmake scripts.

### DIFF
--- a/config/nrfconnect/chip-module/generate_factory_data.cmake
+++ b/config/nrfconnect/chip-module/generate_factory_data.cmake
@@ -180,7 +180,7 @@ endfunction()
 #
 function(nrfconnect_generate_factory_data)
 
-find_package(Python REQUIRED)
+find_package(Python3 REQUIRED)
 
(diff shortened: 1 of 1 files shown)


markaj-nordic (approved):

jangalda-nsc (approved):

config/nrfconnect/chip-module/generate_factory_data.cmake
```diff
@@ -180,7 +180,7 @@ endfunction()
 #
 function(nrfconnect_generate_factory_data)
 
-find_package(Python REQUIRED)
+find_package(Python3 REQUIRED)
```
LuDuda, line 183: Does it work under windows and toolchain manager? :) Let's try..

ArekBalysNordic: It should work since all other NCS Cmake scripts call it in this way, but I would appreciate it if you test it. :)

PR #317, opened 2023-08-16
[shell] Add commands for resource usage statistics

Damian-Nordic:
Add shell commands for printing and resetting the top resource usage statistics. If the heap usage watermark is supported, the top heap usage is printed as well.


config/zephyr/Kconfig
```diff
@@ -281,6 +281,12 @@ config CHIP_PERSISTENT_SUBSCRIPTIONS
 	  subscriptions instead of waiting for the subscriber node to realize that
 	  the publisher is alive again.
 
+config CHIP_STATISTICS
```
kkasperczyk-no, line 284: Shouldn't we add depends on CHIP_LIB_SHELL here? Generally I think we should add information that this is shell module, as based on Kconfig name and description it seems to be a generic module, while it cannot be used outside of the shell, right?

config/nrfconnect/chip-module/CMakeLists.txt
```diff
@@ -130,6 +130,7 @@ matter_add_gn_arg_bool  ("chip_detail_logging"                    CONFIG_MATTER_
 matter_add_gn_arg_bool  ("chip_automation_logging"                FALSE)
 matter_add_gn_arg_bool  ("chip_malloc_sys_heap"                   CONFIG_CHIP_MALLOC_SYS_HEAP)
 matter_add_gn_arg_bool  ("chip_enable_wifi"                       CONFIG_WIFI_NRF700X)
+matter_add_gn_arg_bool  ("chip_system_config_provide_statistics"  CONFIG_CHIP_STATISTICS)
```
kkasperczyk-no, line 133: Similarly as in the comment above, I would consider including shell in name and maybe making it shorter, like `chip_system_statistics_shell`.

kkasperczyk-no (approved):
Left some minor comments.

config/zephyr/Kconfig
Damian-Nordic: In fact, it can. A user might want to enable statistics and use a 3rd-party channel to retrieve them.

kkasperczyk-no: I mean that current implementation doesn't expose API to get statistics information from some 3rd party module, as the implementation is invoked from command handlers.

Damian-Nordic: So this Kconfig switch only enables macros defined in `SystemStats.h` to make sure that `System::Stats::GetResourcesInUse` and `System::Stats::GetHighWatermarks` return meaningful data. Shell is one of the consumers of these values, but one enable `CONFIG_CHIP_STATISTICS` **and** not enable shell at all, and e.g. add a custom BLE service that would expose numbers from the mentioned functions. That's why I think this configuration does not depend on the shell. Does that make sense?

kkasperczyk-no: Ah I see! I missed the fact that gn flag is mapped on the `CHIP_SYSTEM_CONFIG_PROVIDE_STATISTICS` and I was looking for `CHIP_STATISTICS` uses. Yes, now it makes sense.

LuDuda (approved):
This is awesome!

PR #324, opened 2023-09-08
[nrf noup] Enhance/Optimize attach to thread network

doublemis1:
Enhance attaching to Thread by verifying if the current dataset is the same as provided. 
In current solution implementation will not reset the Thread interface while the 
dataset are equal to the active dataset.

### DIFF
--- a/src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.cpp
+++ b/src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.cpp
@@ -353,6 +353,14 @@ template <class ImplClass>
 CHIP_ERROR GenericThreadStackManagerImpl_OpenThread<ImplClass>::_AttachToThreadNetwork(
     const Thread::OperationalDataset & dataset, NetworkCommissioning::Internal::WirelessDriver::ConnectCallback * callback)
 {
+    Thread::OperationalDataset current_dataset;
+    // Validate the dataset change with the current state
+    ThreadStackMgrImpl().GetThreadProvision(current_dataset);
+    if (dataset.AsByteSpan().data_equal(current_dataset.AsByteSpan()) && callback == nullptr)
+    {
+       return CHIP_NO_ERROR;
+    }
+    
     // Reset the previously set callback since it will never be called in case incorrect dataset was supplied.
     mpConnectCallback = nullptr;
     ReturnErrorOnFailure(Impl()->SetThreadEnabled(false));


src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.cpp
```diff
@@ -353,11 +353,16 @@ template <class ImplClass>
...
 {
-    // Reset the previously set callback since it will never be called in case incorrect dataset was supplied.
-    mpConnectCallback = nullptr;
-    ReturnErrorOnFailure(Impl()->SetThreadEnabled(false));
-    ReturnErrorOnFailure(Impl()->SetThreadProvision(dataset.AsByteSpan()));
-
+    Thread::OperationalDataset current_dataset;
+    // Validation the dataset change with the current state
+    ThreadStackMgrImpl().GetThreadProvision(current_dataset);
+    if (!dataset.AsByteSpan().data_equal(current_dataset.AsByteSpan()))
```
Damian-Nordic, line 359: Hmm, maybe you could just do sth like:
```
if (dataset.AsByteSpan().data_equal(current_dataset.AsByteSpan() && callback == nullptr)
{
   return CHIP_NO_ERROR;
}
```
1. If the callback is provided, we probably want to restart the interface. Otherwise, the callback will never be called, so if a user called `ConnectNetwork` for the second time, it would fail (I don't know why a user would want to do it, but it should probably work).
2. If the callback is not provided we don't need to execute `if (dataset.IsCommissioned())` code.

Damian-Nordic (approved):
A small suggestion to reduce the impact of the change.

doublemis1:
Tested after update on NCS 2.2.0 Door Lock NRF52840:
How we can provide fix to clients (e.g. NCS 2.2.0)?

kkasperczyk-no:
> Tested after update on NCS 2.2.0 Door Lock NRF52840: How we can provide fix to clients (e.g. NCS 2.2.0)?

Would creating a known issue with a recommendation to cherry-pick this commit work? Btw. could you open a PR to sdk-nrf with manifest change pointing to this PR? We need to test it against NCS main to be able to merge it.

kkasperczyk-no (approved):

src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.cpp
```diff
@@ -353,6 +353,14 @@ template <class ImplClass>
 CHIP_ERROR GenericThreadStackManagerImpl_OpenThread<ImplClass>::_AttachToThreadNetwork(
     const Thread::OperationalDataset & dataset, NetworkCommissioning::Internal::WirelessDriver::ConnectCallback * callback)
 {
+    Thread::OperationalDataset current_dataset;
+    // Validation the dataset change with the current state
```
markaj-nordic, line 357:
```suggestion
    // Validate the dataset change with the current state
```

markaj-nordic (approved):
@doublemis1 Please open the sdk-nrf PR with bumped manifest.

PR #325, opened 2023-09-13
[nrf noup] Remove recovery mechanism for Wi-Fi connection

ArekBalysNordic:
Currently, the Wi-Fi driver supports the recovery mechanism and it seems that it works properly so we can remove our recovery mechanism.

### DIFF
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -142,8 +142,8 @@ const Map<uint32_t, WiFiManager::NetEventHandler, 5>
     WiFiManager::sEventHandlerMap({ { NET_EVENT_WIFI_SCAN_RESULT, WiFiManager::ScanResultHandler },
                                     { NET_EVENT_WIFI_SCAN_DONE, WiFiManager::ScanDoneHandler },
                                     { NET_EVENT_WIFI_CONNECT_RESULT, WiFiManager::ConnectHandler },
-                                    { NET_EVENT_WIFI_DISCONNECT_RESULT, WiFiManager::NetworkDrivenDisconnectHandler },
-                                    { NET_EVENT_WIFI_DISCONNECT_COMPLETE, WiFiManager::ApplicationDrivenDisconnectHandler } });
+                                    { NET_EVENT_WIFI_DISCONNECT_RESULT, WiFiManager::DisconnectHandler }, 
+                                    { NET_EVENT_WIFI_DISCONNECT_COMPLETE, WiFiManager::DisconnectHandler } });
 
 void WiFiManager::WifiMgmtEventHandler(net_mgmt_event_callback * cb, uint32_t mgmtEvent, net_if * iface)
 {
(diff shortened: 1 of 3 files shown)


kkasperczyk-no (approved):

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -557,20 +511,7 @@ void WiFiManager::NetworkDrivenDisconnectHandler(Platform::UniquePtr<uint8_t>)
 
 void WiFiManager::ApplicationDrivenDisconnectHandler(Platform::UniquePtr<uint8_t>)
 {
-    if (!Instance().mRecoveryArmed)
```
markaj-nordic, line 560: I think we should verify and better understand the behavior of the `NET_EVENT_WIFI_DISCONNECT_RESULT` and `NET_EVENT_WIFI_DISCONNECT_COMPLETE` events. Judging by the naming, the former just forwards if the disconnection request was successful and the latter indicates if the station was really disconnected from the AP. Yet, we used these two to distinguish between self-disconnects and disconnections forced by the application code with Instance().Disconnect(). The main reason why I raise this concern is that we need to know exactly when to inform the application that the station is disconnected with the following code:

```
    SystemLayer().ScheduleLambda([] {
        ChipLogProgress(DeviceLayer, "WiFi station disconnected");
        Instance().mWiFiState = WIFI_STATE_DISCONNECTED;
        Instance().PostConnectivityStatusChange(kConnectivity_Lost);
    });
```
which currently is placed in the `NetworkDrivenDisconnectHandler` but maybe it should be moved or copied here (depending on the relation between these two events).

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -206,32 +206,12 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
     mCachedWiFiState    = mWiFiState;
     mWiFiState          = WIFI_STATE_SCANNING;
     mSsidFound          = false;
-    mRecoveryArmed      = true;
-    // TODO Workaround for recovery mechanism to wait before the next scan request until the WiFi supplicant is not busy.
-    static bool workaroundDone;
 
     int ret = net_mgmt(NET_REQUEST_WIFI_SCAN, iface, NULL, 0);
```
markaj-nordic, line 210: nit: I guess we don't need the temporary variable now.

src/platform/nrfconnect/wifi/WiFiManager.cpp
ArekBalysNordic: I can see that in the supplicant/driver code and the examples, they rely on `NET_EVENT_WIFI_DISCONNECT_RESULT` only. They use it to inform the upper layer that the connection has been closed. So in our implementation, after removing the workarounds, we don't need to know that the request comes from the application or network and we don't need to support both events anymore, we can leave only `NET_EVENT_WIFI_DISCONNECT_RESULT` and assign it to the disconnection callback. I will remove it properly.

krish2718: The `NET_EVENT_WIFI_DISCONNECT_COMPLETE` was added to ACK a disconnect request from the application saying that disconnection is now completed. `NET_EVENT_WIFI_DISCONNECT_RESULT` is a unsolicited notification that the connection no longer exists.

 

* App requests `NET_REQUEST_WIFI_DISCONNECT`
    - Once WPA supplicant completes disconnect it send `NET_EVENT_WIFI_DISCONNECT_COMPLETE`
* App requests connection and is success
    - AP disconnects the STA, then WPA supplicant sends `NET_EVENT_WIFI_DISCONNECT_RESULT`

We should probably rename `NET_EVENT_WIFI_DISCONNECT_RESULT` -> `NET_EVENT_WIFI_DISCONNECTED` or something.

ArekBalysNordic: @krish2718 Thank you for the explanation! So @markaj-nordic, as I understood correctly, now, in our case, as we don't need to know if the Application wanted to disconnect or disconnection came from a different source, we should use only the NET_EVENT_WIFI_DISCONNECT_RESULT event to propagate it to the Application layer. Right?

markaj-nordic: I'd say that we may want to keep both handlers because the `WiFiManager::Disconnect` is a part of public API and `NrfWiFiDriver` may call the  when reverting the configuration in the fail-safe mechanism. So in `NET_EVENT_WIFI_DISCONNECT_COMPLETE` handler we would simply have the code snippet from my first comment.

ArekBalysNordic: Ok, so maybe we can just assign both events to the same callback, due to the fact that both of the callbacks will do exactly the same things?

markaj-nordic: Sure, if the code is the same that would be perfect.

markaj-nordic (approved):

PR #331, opened 2023-10-09
[crypto] Add PSA SPAKE2+ implementation

Damian-Nordic:
Implement Matter's SPAKE2+ protocol using PSA crypto API. Use the PSA implementation in PASE session for nRF Connect platform.

**Note:** The implementation can't be used without changes in PSA crypto driver listed in NCSDK-23835.

### DIFF
--- a/src/crypto/PSASpake2p.cpp
+++ b/src/crypto/PSASpake2p.cpp
@@ -1,0 +1,203 @@
+/*
+ *    Copyright (c) 2023 Project CHIP Authors
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#include "PSASpake2p.h"
+
+#include "CHIPCryptoPALPSA.h"
+
+#include <lib/support/CodeUtils.h>
+
+#include <psa/crypto.h>
+
+namespace chip {
+namespace Crypto {
+
+CHIP_ERROR PSASpake2p_P256_SHA256_HKDF_HMAC::Init(const uint8_t * context, size_t context_len)
+{
+    Clear();
+
+    VerifyOrReturnError(context_len <= sizeof(mContext), CHIP_ERROR_BUFFER_TOO_SMALL);
+
+    psa_pake_cipher_suite_t cs = PSA_PAKE_CIPHER_SUITE_INIT;
+    psa_pake_cs_set_algorithm(&cs, PSA_ALG_SPAKE2P);
+    psa_pake_cs_set_primitive(&cs, PSA_PAKE_PRIMITIVE(PSA_PAKE_PRIMITIVE_TYPE_ECC, PSA_ECC_FAMILY_SECP_R1, 256));
+    psa_pake_cs_set_hash(&cs, PSA_ALG_SHA_256);
+
+    psa_status_t status = psa_pake_setup(&mOperation, &cs);
+    VerifyOrReturnError(status == PSA_SUCCESS, CHIP_ERROR_INTERNAL);
+
+    memcpy(mContext, context, context_len);
+    mContextLen = context_len;
+
+    return CHIP_NO_ERROR;
+}
+
+void PSASpake2p_P256_SHA256_HKDF_HMAC::Clear()
+{
+    IgnoreUnusedVariable(psa_pake_abort(&mOperation));
+    mOperation = psa_pake_operation_init();
+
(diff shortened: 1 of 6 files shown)


src/crypto/PSASpake2p.h
```diff
@@ -0,0 +1,156 @@
...
+     * @brief Compute the second round of the protocol.
+     *
+     * @param in      The peer first round Spake2+ contribution.
+     * @param in_len  The peer first round Spake2+ contribution length.
+     * @param out     The output second round Spake2+ contribution.
+     * @param out_len The output second round Spake2+ contribution length.
+     *
+     * @return Returns a CHIP_ERROR on error, CHIP_NO_ERROR otherwise
+     **/
+    CHIP_ERROR ComputeRoundTwo(const uint8_t * in, size_t in_len, uint8_t * out, size_t * out_len);
```
kkasperczyk-no, line 124: Is `out_len` in-out parameter? Looking at the implementation it seems to me the `out_len` should have a valid value determining the max allowed size of `out` buffer and on return it will be filled with the number of bytes returned. 
If this is correct maybe it would not harm to add such information to the docs (or split max size and output length into 2 parameters)

src/crypto/PSASpake2p.cpp
```diff
@@ -0,0 +1,197 @@
...
+void PSASpake2p_P256_SHA256_HKDF_HMAC::Clear()
+{
+    IgnoreUnusedVariable(psa_pake_abort(&mOperation));
+    mOperation = psa_pake_operation_init();
+
+    IgnoreUnusedVariable(psa_destroy_key(mKey));
+    mKey = PSA_KEY_ID_NULL;
+}
+
+CHIP_ERROR PSASpake2p_P256_SHA256_HKDF_HMAC::BeginVerifier(const uint8_t * my_identity, size_t my_identity_len,
```
kkasperczyk-no, line 58: I observed that here and in many other places there aren't nullptr checks for input parameters. Is it assumed that the null checks are done one level lower (by the psa_... methods implementation)?

src/crypto/PSASpake2p.cpp
```diff
@@ -0,0 +1,197 @@
...
+CHIP_ERROR PSASpake2p_P256_SHA256_HKDF_HMAC::GetKeys(uint8_t * out, size_t * out_len)
+{
+    /*
+     * TODO: either:
+     * - use psa_pake_shared_secret() proposed in https://github.com/ARM-software/psa-api/issues/86
+     * - refactor Matter's GetKeys API to take an abstract shared secret instead of raw secret bytes.
+     */
+    oberon_spake2p_operation_t & oberonCtx = mOperation.MBEDTLS_PRIVATE(ctx).oberon_spake2p_ctx;
+
+    VerifyOrReturnError(out_len != nullptr, CHIP_ERROR_INVALID_ARGUMENT);
```
kkasperczyk-no, line 187: Shouldn't we verify `out != nullptr` as well?

src/crypto/PSASpake2p.cpp
```diff
@@ -0,0 +1,197 @@
...
+#include <psa/crypto.h>
+
+namespace chip {
+namespace Crypto {
+
+CHIP_ERROR PSASpake2p_P256_SHA256_HKDF_HMAC::Init(const uint8_t * context, size_t context_len)
+{
+    Clear();
+
+    VerifyOrReturnError(context_len <= sizeof(mContext), CHIP_ERROR_INVALID_ARGUMENT);
```
kkasperczyk-no, line 33: Wouldn't `CHIP_ERROR_BUFFER_TOO_SMALL` be more appropriate?

src/crypto/PSASpake2p.cpp
```diff
@@ -0,0 +1,197 @@
...
+{
+    psa_status_t status;
+
+    if (mRole == PSA_PAKE_ROLE_CLIENT)
+    {
+        status = psa_pake_input(&mOperation, PSA_PAKE_STEP_KEY_SHARE, in, in_len);
+        VerifyOrReturnError(status == PSA_SUCCESS, CHIP_ERROR_INTERNAL);
+    }
+
+    status = psa_pake_output(&mOperation, PSA_PAKE_STEP_CONFIRM, out, *out_len, out_len);
```
kkasperczyk-no, line 164: Are you sure it's safe to read `*out_len` value from pointer that could be potentially nullptr? I think it can lead to usage fault, right?

src/crypto/PSASpake2p.h
Damian-Nordic: This interface mimics existing Spake2p class which indeed treats out_len as in-out parameter. But I can extend the docs, sure.

src/crypto/PSASpake2p.cpp
Damian-Nordic: Yes, PSA driver checks if the pointers are not null.

src/crypto/PSASpake2p.cpp
Damian-Nordic: Yeah, can be. I'll change.

src/crypto/PSASpake2p.cpp
Damian-Nordic: Added extra checks.

kkasperczyk-no (approved):

ArekBalysNordic (approved):
LGTM

LuDuda (approved):
This looks great! Thanks :+1:

PR #332, opened 2023-10-09
[nrf fromtree] app: clusters: Add handling a local lock operation error

ArekBalysNordic:
In case we use a custom handling lock credential we need a mechanism that allows us to handle a lock operation error when the wrong credential has been provided.

Added handling local lock operation error to emit LockOperationEvent and call HandleWrongCodeEntry.

### DIFF
--- a/src/app/clusters/door-lock-server/door-lock-server.cpp
+++ b/src/app/clusters/door-lock-server/door-lock-server.cpp
@@ -188,6 +188,18 @@ bool DoorLockServer::SetPrivacyModeButton(chip::EndpointId endpointId, bool isEn
     return SetAttribute(endpointId, Attributes::EnablePrivacyModeButton::Id, Attributes::EnablePrivacyModeButton::Set, isEnabled);
 }
 
+void DoorLockServer::HandleLocalLockOperationError(chip::EndpointId endpointId, LockOperationTypeEnum opType,
+                                                   OperationSourceEnum opSource, Nullable<uint16_t> userId)
+{
+    SendLockOperationEvent(endpointId, opType, opSource, OperationErrorEnum::kInvalidCredential, userId,
+                           Nullable<chip::FabricIndex>(), Nullable<chip::NodeId>(), Nullable<List<const LockOpCredentials>>(),
+                           false);
+
+    HandleWrongCodeEntry(endpointId);
+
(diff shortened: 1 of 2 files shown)


markaj-nordic:
Strange, I thought this had been addressed in the Matter upstream.

src/app/clusters/door-lock-server/door-lock-server.cpp
```diff
@@ -188,6 +188,18 @@ bool DoorLockServer::SetPrivacyModeButton(chip::EndpointId endpointId, bool isEn
     return SetAttribute(endpointId, Attributes::EnablePrivacyModeButton::Id, Attributes::EnablePrivacyModeButton::Set, isEnabled);
 }
 
+void DoorLockServer::HandleLocalLockOperationError(chip::EndpointId endpointId, LockOperationTypeEnum opType,
+                                                   OperationSourceEnum opSource, Nullable<uint16_t> userId)
+{
+    SendLockOperationEvent(endpointId, opType, opSource, OperationErrorEnum::kInvalidCredential, userId,
```
markaj-nordic, line 194: Is it enough to pass fabric idx and node id always as nulls? I mean is it compliant with the spec?

ArekBalysNordic:
> Strange, I thought this had been addressed in the Matter upstream.

Should be, but actually isn't. I didn't find another way to handle it, only the remote operation is handled within the Lock server and the SendLockOperationEvent method is private so I guess it shouldn't be used outside the Lock server. We can upstream it also if you want, or don't do anything until it is addressed in Matter upstream.

src/app/clusters/door-lock-server/door-lock-server.cpp
ArekBalysNordic: This is a local operation, so we don't have any fabric, or node id, just an endpoint.

markaj-nordic (approved):

ArekBalysNordic:
> I would recommend upstreaming this change, unless we are sure we can remember about it during upmerging :)
> 
> Why this is targeted only to v2.4-branch?

We need it only for FELL purposes for now, and this project includes the v2.4-branch. I will upstream those changes and we will have it in the sdk-connectedhomeip main branch later.

PR #335, opened 2023-10-10
[nrf fromlist] Clear endpoint bitmask before reporting the PartList attribute

ArekBalysNordic:
When we set a low value to the minimum subscription interval for subscription for the PartList attribute of the Description cluster, and then remove a dynamic endpoint the related bitmask may change after reporting it to a controller.
The controller receives the old number of endpoints and does not update it within the following subscriptions until the next change of the PartList attribute occurs.

To resolve the issue the endpoint bitmask should be cleared before reporting the PartList attribute of the Descriptor cluster to avoid race conditions.

[nrf] -> There is an old API and there is a bitmask operation instead of the method Clear!

### DIFF
--- a/src/app/reporting/Engine.cpp
+++ b/src/app/reporting/Engine.cpp
@@ -829,7 +829,7 @@ CHIP_ERROR Engine::SetDirty(AttributePathParams & aAttributePath)
 
     bool intersectsInterestPath = false;
     InteractionModelEngine::GetInstance()->mReadHandlers.ForEachActiveObject(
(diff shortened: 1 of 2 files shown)


LuDuda (approved):

kkasperczyk-no (approved):
LGTM, but could you explicitly comment in the code the part that is a workaround, as "TODO: remove this workaround after upmerge to Matter 1.2"? I would like to just be sure we will not forget to do that.

PR #342, opened 2023-10-19
[nrf noup] config: ncs: Fix Wi-Fi maximum transmit tokens

krish2718:
Wi-Fi samples use 10 as tokens, i.e., no spare tokens at all (5 * 2 =
    10), but Matter samples use 8 which gives ( 5 * 1 + 3), so, 3 spare
    tokens which changes the traffic profile.
    
    So, match with Wi-Fi as that combination is well tested.


krish2718:
subset of #341, so, closing.

PR #357, opened 2023-11-23
[nrf noup] [nrfconnect] Move handler for joining/leaving mcast group

Damian-Nordic:
Move the platform handler for joining and leaving a multicast group to ConnectivityManagerImpl to support Matter stack on a system with multiple network interfaces (Thread + Wi-Fi). Previously, both Thread and WiFi would try to register their handler but only one of them would win.

### DIFF
--- a/src/platform/nrfconnect/ConnectivityManagerImpl.cpp
+++ b/src/platform/nrfconnect/ConnectivityManagerImpl.cpp
@@ -17,11 +17,12 @@
 
 #include <platform/internal/CHIPDeviceLayerInternal.h>
 
-#include <platform/ConnectivityManager.h>
-#include <platform/internal/BLEManager.h>
-
+#include <inet/UDPEndPointImplSockets.h>
 #include <lib/support/CodeUtils.h>
 #include <lib/support/logging/CHIPLogging.h>
+#include <platform/ConnectivityManager.h>
+#include <platform/internal/BLEManager.h>
+#include <platform/Zephyr/InetUtils.h>
 
 #include <platform/internal/GenericConnectivityManagerImpl_UDP.ipp>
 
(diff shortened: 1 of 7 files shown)


LuDuda (approved):
Thanks 👍

markaj-nordic (approved):
LGTM

src/platform/nrfconnect/ConnectivityManagerImpl.cpp
```diff
@@ -35,15 +36,55 @@
...
+        otError error;
+
+        ThreadStackMgr().LockThreadStack();
+        error = handler(openthread_get_default_instance(), &otAddress);
+        ThreadStackMgr().UnlockThreadStack();
+
+        return MapOpenThreadError(error);
+    }
+#endif
+
```
markaj-nordic, line 65: Is there any chance that we have L2 different than WIFI and we still end up in the code above? I mean, maybe we can just similarly:

`if (net_if_l2(iface) == &NET_L2_GET_NAME(WIFI))` *(not sure if this work out this way)

before registering multicasts.

Damian-Nordic: I had that in the previous version but then I thought that the same code could be used not only for Wi-Fi but also for Ethernet or other interfaces one could think of so the condition wouldn't add any value. But given that it's under `ifdef` anyway, I can re-add the condition  to make it less confusing.

markaj-nordic: I see, so it's up to you then. Maybe a comment that the same code is valid for Ethernet is enough.

Damian-Nordic: ok, I added a comment.

PR #360, opened 2023-11-24
[nrf noup] Clean Kconfig.defaults to support "Thread and Wi-Fi" builds

Damian-Nordic:
1. Group Kconfig options into categories to make it easier to see discrepancies between Thread and Wi-Fi and help choose the right value when both transports are enabled.
    
   That is, prefer:
    
       config A
         default X if CHIP_WIFI
         default Y if NET_L2_OPENTHREAD
    
   over:
    
       if CHIP_WIFI
       config A
         default X
       endif
    
       if NET_L2_OPENTHREAD
       config A
         default Y
       endif
2. Remove redundant defaults that are already selected in other Kconfig files.
3. Enable SHA1 for Wi-Fi

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig.defaults
+++ b/config/nrfconnect/chip-module/Kconfig.defaults
@@ -304,6 +304,9 @@ config MBEDTLS_CTR_DRBG_C
 config MBEDTLS_CIPHER_MODE_CTR
     default y
 
+config MBEDTLS_SHA1_C
+    default y if CHIP_WIFI
+
 config MBEDTLS_SHA256_C
     default y
 


LuDuda (approved):
This is fantastic! Thank you so much!

Just a question: Are we able to easily check (other than in manual way) if some of the configuration in Kconfig.defaults are not necessary anymore? For example with introduction of `SHELL_MINIMAL` i do see now, that i could have removed `KERNEL_SHELL`  config. I bet, some other configs are not needed anymore as well.

kkasperczyk-no (approved):

Damian-Nordic:
> This is fantastic! Thank you so much!
> 
> Just a question: Are we able to easily check (other than in manual way) if some of the configuration in Kconfig.defaults are not necessary anymore? For example with introduction of `SHELL_MINIMAL` i do see now, that i could have removed `KERNEL_SHELL` config. I bet, some other configs are not needed anymore as well.

I don't know, we would need to write a script that removes random configs and checks if `.config` hasn't changed, but I'm not aware of such a tool.

PR #364, opened 2023-12-05
[nrf noup] Added check to drop handling callbacks for BT central

kkasperczyk-no:
Matter BLEManager handles all BT connect and disconnect callbacks no matter if these are Matter related ones or not. It collides with other not Matter-related services that trigger Matter CHIPoBLE service advertising changes. Added role check that allows to at least drop all callbacks related to BT central role.

### DIFF
--- a/src/platform/Zephyr/BLEManagerImpl.cpp
+++ b/src/platform/Zephyr/BLEManagerImpl.cpp
@@ -146,7 +146,8 @@ CHIP_ERROR BLEManagerImpl::_Init()
     mServiceMode = ConnectivityManager::kCHIPoBLEServiceMode_Enabled;
     mFlags.ClearAll().Set(Flags::kAdvertisingEnabled, CHIP_DEVICE_CONFIG_CHIPOBLE_ENABLE_ADVERTISING_AUTOSTART);
     mFlags.Set(Flags::kFastAdvertisingEnabled, true);
-    mGAPConns = 0;
+    mMatterConnNum = 0;
+    mTotalConnNum  = 0;
 
     memset(mSubscribedConns, 0, sizeof(mSubscribedConns));
 
@@ -409,14 +410,12 @@ CHIP_ERROR BLEManagerImpl::HandleGAPConnect(const ChipDeviceEvent * event)
     if (connEvent->HciResult == BT_HCI_ERR_SUCCESS)
     {
         ChipLogProgress(DeviceLayer, "BLE connection established (ConnId: 0x%02x)", bt_conn_index(connEvent->BtConn));
-        mGAPConns++;
+        mMatterConnNum++;
     }
     else
     {
         ChipLogError(DeviceLayer, "BLE connection failed (reason: 0x%02x)", connEvent->HciResult);
     }
-
-    ChipLogProgress(DeviceLayer, "Current number of connections: %u/%u", NumConnections(), CONFIG_BT_MAX_CONN);
 
     mFlags.Set(Flags::kAdvertisingRefreshNeeded);
     PlatformMgr().ScheduleWork(DriveBLEState, 0);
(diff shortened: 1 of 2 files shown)


src/platform/Zephyr/BLEManagerImpl.h
```diff
@@ -138,6 +138,9 @@ class BLEManagerImpl final : public BLEManager, private BleLayer, private BlePla
 
     static BLEManagerImpl sInstance;
 
+    // The summarized number of Bluetooth LE connections related to the device (including these not related to Matter service).
+    static uint16_t sBleConnNum;
```
markaj-nordic, line 142: Is this only used in logging of the current number of connections? BTW. The `static` is probably not needed here as the whole class is implemented as a singleton (we could spare additional definition in cpp file).

src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -790,10 +789,16 @@ void BLEManagerImpl::HandleTXIndicated(struct bt_conn * conId, bt_gatt_indicate_
 void BLEManagerImpl::HandleConnect(struct bt_conn * conId, uint8_t err)
 {
     ChipDeviceEvent event;
+    struct bt_conn_info bt_info;
```
markaj-nordic, line 792: `struct` not needed, also in other places.

src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -790,10 +789,16 @@ void BLEManagerImpl::HandleTXIndicated(struct bt_conn * conId, bt_gatt_indicate_
...
 
     PlatformMgr().LockChipStack();
 
+    sBleConnNum++;
+    ChipLogProgress(DeviceLayer, "Current number of connections: %u/%u", sBleConnNum, CONFIG_BT_MAX_CONN);
+
     // Don't handle BLE connecting events when it is not related to CHIPoBLE
+    VerifyOrExit(bt_conn_get_info(conId, &bt_info) == 0, );
+    VerifyOrExit(bt_info.role == BT_CONN_ROLE_PERIPHERAL, );
```
markaj-nordic, line 801: I'd suggest to add a comment why we drop non-peripheral connections.

markaj-nordic (approved):

src/platform/Zephyr/BLEManagerImpl.h
kkasperczyk-no: Yes, initially I was going to remove this log, but after all it seems to be useful. Unfortunately I couldn't find the way to get from BT stack the current number of connected devices, so I had to create a counter.

src/platform/Zephyr/BLEManagerImpl.cpp
kkasperczyk-no: Ok, I thought it's clear based on the comment above (Don't handle BLE connecting events when it is not related to CHIPoBLE) but I can further clarify that according to Matter spec we have to be a peripheral.

src/platform/Zephyr/BLEManagerImpl.h
```diff
@@ -105,6 +105,8 @@ class BLEManagerImpl final : public BLEManager, private BleLayer, private BlePla
 #if CHIP_ENABLE_ADDITIONAL_DATA_ADVERTISING
     PacketBufferHandle c3CharDataBufferHandle;
 #endif
+    // The summarized number of Bluetooth LE connections related to the device (including these not related to Matter service).
+    uint16_t mBleConnNum;
```
Damian-Nordic, line 109: nit: could you better name `mGAPConns` to distinguish it from `mBleConnNum`. Maybe `mMatterConnNum` vs `mTotalConnNum`?

Damian-Nordic (approved):

ArekBalysNordic (approved):

PR #370, opened 2023-12-19
[nrf noup] Versioning with VERSION file for matter implemented

LipinskiPNordicSemi:
This commit adds possibility to use VERSION file which will work properly for DFU SMP and Matter OTA.

Users still can use CHIP_DEVICE_SOFTWARE_VERSION or MCUBOOT_IMGTOOL_SIGN_VERSION config.

These changes disable Kconfig for Matter OTA values only when VERSION file is created in the app directory.


config/zephyr/Kconfig
```diff
@@ -95,6 +95,7 @@ config CHIP_DEVICE_HARDWARE_VERSION_STRING
 	  hardware version string is exposed as an attribute of the Basic
 	  Information cluster.
 
+if "$(APPVERSION)"=""
```
Damian-Nordic, line 98: Could you update the PR description with the desired behavior? Do we want the VERSION file to take precedence over CHIP_DEVICE_SOFTWARE_VERSION? If so, what should a customer do in case they already use a different versioning scheme than ${APPVERSION}?

LipinskiPNordicSemi: Good point, thanks. User can still use CHIP_DEVICE_SOFTWARE_VERSION, we wanted to keep backward compability. Only when VERSION file is created the kconfig are disabled.

kkasperczyk-no (approved):
It's looks pretty good to me, but let's wait with merging it, until the samples in the corresponding PR to sdk-nrf will be aligned. I would like to make sure that CI will pass once we start using VERSION scheme.

Damian-Nordic (approved):

ArekBalysNordic (approved):

config/zephyr/Kconfig
LuDuda, line 98: Just a thought.. what about to use `depends` attribute to below KConfig instead of if-statement? In this case user can get warning during building when using both versioning methods. Or its worlaround to fix other issue? ;)

LuDuda (approved):

config/zephyr/Kconfig
LipinskiPNordicSemi: I think that "if" block is change into `depends` statement for every config inside, during the build process.

ArekBalysNordic:
Rebased.

PR #372, opened 2023-12-29
[nrf noup] Move DAC priv key from Factory Data to PSA ITS

ArekBalysNordic:
We need a mechanism to move the DAC private key from the factory data set to PSA ITS NVM storage during the first boot of the device.
Then the DAC private key must be removed from the factory data set and protected by overwriting.

In this commit:
- Added a method to FactoryDataProvider for moving and removing DAC from the factory data set.
- Aligned the SignWithDeviceAttestationKey method to work with stored DAC private key in ITS NVM instead of raw data.
- Extended the FactoryDataParser module by the Serialize method that is responsible for creating CBOR-formated factory data from the FactoryData struct.
- Added a Kconfig to define the PSA ITS NVM offset for Matter keys.

- x509 MBedTLS support seems to be not needed anymore - we can disable it and save ~20kB of FLASH.

- Prevent the DAC private key from removal during the factory reset - for now, disable the CHIP_FACTORY_RESET_ERASE_NVS config by default.

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig
+++ b/config/nrfconnect/chip-module/Kconfig
@@ -107,10 +107,6 @@ config CHIP_FACTORY_DATA
 	bool "Factory data provider"
 	select ZCBOR
 	imply FPROTECT
-	imply MBEDTLS_X509_LIBRARY if CHIP_CRYPTO_PSA
-	imply MBEDTLS_X509_CRT_PARSE_C if CHIP_CRYPTO_PSA
-	imply MBEDTLS_PK_PARSE_C if CHIP_CRYPTO_PSA
-	imply MBEDTLS_TLS_LIBRARY if CHIP_CRYPTO_PSA
 	help
 	  Enables the default nRF Connect factory data provider implementation that
 	  supports reading the factory data encoded in the CBOR format from the
@@ -250,6 +246,7 @@ endif # CHIP_FACTORY_DATA_BUILD
 config CHIP_FACTORY_RESET_ERASE_NVS
 	bool
 	default y
+	depends on !CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
 
 config CHIP_LOG_SIZE_OPTIMIZATION
 	bool "Disable some detailed logs to decrease flash usage"
@@ -283,5 +280,17 @@ config CHIP_ENABLE_READ_CLIENT
 	  Enable support for Read Client in the Interaction Model.
 	  This config can be disabled for device types that do not require Read Client functionality.
 	  Disabling this config can save flash and RAM space.
+
+config CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
+	bool "Migrate DAC private key from factory data to PSA ITS"
+	depends on CHIP_CRYPTO_PSA
+	depends on CHIP_FACTORY_DATA
+	select EXPERIMENTAL
+	help
+	  Move DAC private key from the factory data set to the PSA ITS secure storage
+	  and remove it. After the first boot of the device the DAC private key will be moved
+	  to the PSA ITS secure storage and will not be available in the factory data anymore.
+	  It will be overwritten in the factory data set by zeros.
+	
 
 endif # CHIP
--- a/src/crypto/CHIPCryptoPALPSA.h
+++ b/src/crypto/CHIPCryptoPALPSA.h
@@ -62,7 +62,8 @@ enum class KeyIdBase : psa_key_id_t
 {
     Minimum     = CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE,
     Operational = Minimum, ///< Base of the PSA key ID range for Node Operational Certificate private keys
-    Maximum     = Operational + kMaxValidFabricIndex,
+    DACPrivKey  = Operational + kMaxValidFabricIndex + 1,
+    Maximum     = DACPrivKey,
 };
 
 static_assert(to_underlying(KeyIdBase::Minimum) >= PSA_KEY_ID_USER_MIN && to_underlying(KeyIdBase::Maximum) <= PSA_KEY_ID_USER_MAX,
--- a/src/platform/nrfconnect/CHIPPlatformConfig.h
+++ b/src/platform/nrfconnect/CHIPPlatformConfig.h
@@ -54,6 +54,10 @@
 #endif // CHIP_CONFIG_SHA256_CONTEXT_ALIGN
 #endif // CONFIG_CHIP_CRYPTO_PSA
 
+#ifndef CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE
+#define CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE 0x30000
+#endif // CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE
+
 // ==================== General Configuration Overrides ====================
 
 #ifndef CHIP_CONFIG_MAX_UNSOLICITED_MESSAGE_HANDLERS
--- a/src/platform/nrfconnect/FactoryDataParser.c
+++ b/src/platform/nrfconnect/FactoryDataParser.c
@@ -17,12 +17,13 @@
 
 #include "FactoryDataParser.h"
 
+#include <zcbor_common.h>
 #include <zcbor_decode.h>
 
 #include <ctype.h>
 #include <string.h>
 
-#define MAX_FACTORY_DATA_NESTING_LEVEL 4
+#define MAX_FACTORY_DATA_NESTING_LEVEL 2
 
 static inline bool uint16_decode(zcbor_state_t * states, uint16_t * value)
 {
@@ -123,6 +124,11 @@ bool FindUserDataEntry(struct FactoryData * factoryData, const char * entry, voi
 
 bool ParseFactoryData(uint8_t * buffer, uint16_t bufferSize, struct FactoryData * factoryData)
 {
+    if (!buffer || !factoryData || bufferSize == 0)
+    {
+        return false;
+    }
+
     memset(factoryData, 0, sizeof(*factoryData));
     ZCBOR_STATE_D(states, MAX_FACTORY_DATA_NESTING_LEVEL, buffer, bufferSize, 1);
 
@@ -209,7 +215,8 @@ bool ParseFactoryData(uint8_t * buffer, uint16_t bufferSize, struct FactoryData 
         }
         else if (strncmp("dac_key", (const char *) currentString.value, currentString.len) == 0)
         {
-            res = res && zcbor_bstr_decode(states, (struct zcbor_string *) &factoryData->dac_priv_key);
+            res                              = res && zcbor_bstr_decode(states, (struct zcbor_string *) &factoryData->dac_priv_key);
+            factoryData->dacPrivateKeyOffset = (size_t) ((uint8_t *) factoryData->dac_priv_key.data - buffer);
         }
         else if (strncmp("pai_cert", (const char *) currentString.value, currentString.len) == 0)
         {
--- a/src/platform/nrfconnect/FactoryDataParser.h
+++ b/src/platform/nrfconnect/FactoryDataParser.h
@@ -66,6 +66,7 @@ struct FactoryData
     bool discriminatorPresent;
     bool productFinishPresent;
     bool primaryColorPresent;
+    size_t dacPrivateKeyOffset;
 };
 
 /**
--- a/src/platform/nrfconnect/FactoryDataProvider.cpp
+++ b/src/platform/nrfconnect/FactoryDataProvider.cpp
@@ -24,7 +24,15 @@
 #include <platform/Zephyr/ZephyrConfig.h>
 #endif
 
-#include <zephyr/logging/log.h>
+#include <lib/support/logging/CHIPLogging.h>
+
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+#include <lib/support/ScopedBuffer.h>
+#include <psa/crypto.h>
+#include <zephyr/drivers/flash.h>
+
+static const struct device * const kFlashDev = DEVICE_DT_GET_OR_NULL(DT_CHOSEN(zephyr_flash_controller));
+#endif
 
 namespace chip {
 namespace {
@@ -59,21 +67,7 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
     uint8_t * factoryData = nullptr;
     size_t factoryDataSize;
 
-    CHIP_ERROR error = mFlashFactoryData.ProtectFactoryDataPartitionAgainstWrite();
-
-    // Protection against write for external storage is not supported.
-    if (error == CHIP_ERROR_NOT_IMPLEMENTED)
-    {
-        ChipLogProgress(DeviceLayer, "The device does not support hardware protection against write.");
-        error = CHIP_NO_ERROR;
-    }
-    else if (error != CHIP_NO_ERROR)
-    {
-        ChipLogError(DeviceLayer, "Failed to protect the factory data partition.");
-        return error;
-    }
-
-    error = mFlashFactoryData.GetFactoryDataPartition(factoryData, factoryDataSize);
+    CHIP_ERROR error = mFlashFactoryData.GetFactoryDataPartition(factoryData, factoryDataSize);
 
     if (error != CHIP_NO_ERROR)
     {
@@ -95,8 +89,122 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
         return CHIP_ERROR_VERSION_MISMATCH;
     }
 
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+    VerifyOrDie(MoveDACPrivateKeyToSecureStorage(factoryData, factoryDataSize) == CHIP_NO_ERROR);
+#endif
+
+    error = mFlashFactoryData.ProtectFactoryDataPartitionAgainstWrite();
+
+    // Protection against write for external storage is not supported.
+    if (error == CHIP_ERROR_NOT_IMPLEMENTED)
+    {
+        ChipLogProgress(DeviceLayer, "The device does not support hardware protection against write.");
+        error = CHIP_NO_ERROR;
+    }
+    else if (error != CHIP_NO_ERROR)
+    {
+        ChipLogError(DeviceLayer, "Failed to protect the factory data partition.");
+        return error;
+    }
+
+    return error;
+}
+
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+template <class FlashFactoryData>
+CHIP_ERROR FactoryDataProvider<FlashFactoryData>::MoveDACPrivateKeyToSecureStorage(uint8_t * factoryDataPartition,
+                                                                                   size_t factoryDataSize)
+{
+
+    if (!factoryDataPartition || factoryDataSize == 0)
+    {
+        return CHIP_ERROR_INVALID_ARGUMENT;
+    }
+
+    if (mFactoryData.dac_priv_key.len != kDACPrivateKeyLength)
+    {
+        return CHIP_ERROR_INVALID_LIST_LENGTH;
+    }
+
+    uint8_t clearedDACPrivKey[kDACPrivateKeyLength];
+    memset(clearedDACPrivKey, 0x00, sizeof(clearedDACPrivKey));
+
+    // Check if factory data contains DAC private key
+    if (memcmp(mFactoryData.dac_priv_key.data, clearedDACPrivKey, kDACPrivateKeyLength) != 0)
+    {
+        psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+        // If there is the new DAC private key present in the factory data set and also there is
+        // the existing one in ITS NVM storage, then skip saving it again.
+        if (psa_get_key_attributes(mDACPrivKeyId, &attributes) != PSA_SUCCESS)
+        {
+            ChipLogProgress(DeviceLayer, "Found DAC Private Key in factory data set. Copying to secure storage...");
+
+            // Remove the key if any exists and can be corrupted.
+            psa_destroy_key(mDACPrivKeyId);
+
+            psa_reset_key_attributes(&attributes);
+            psa_set_key_type(&attributes, PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_SECP_R1));
+            psa_set_key_bits(&attributes, kDACPrivateKeyLength * 8);
+            psa_set_key_algorithm(&attributes, PSA_ALG_ECDSA(PSA_ALG_SHA_256));
+#ifdef CONFIG_CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
+            psa_set_key_lifetime(&attributes, PSA_KEY_LIFETIME_PERSISTENT);
+            psa_set_key_id(&attributes, mDACPrivKeyId);
+#else
+            psa_set_key_lifetime(&attributes, PSA_KEY_LIFETIME_VOLATILE);
+#endif
+            psa_set_key_usage_flags(&attributes, PSA_KEY_USAGE_SIGN_MESSAGE);
+
+            VerifyOrReturnError(psa_import_key(&attributes, reinterpret_cast<uint8_t *>(mFactoryData.dac_priv_key.data),
+                                               kDACPrivateKeyLength, &mDACPrivKeyId) == PSA_SUCCESS,
+                                CHIP_ERROR_INTERNAL);
+        }
+
+#ifdef CONFIG_CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
+        // Check once again if the saved key has attributes set before removing it from the factory data set.
+        VerifyOrReturnError(psa_get_key_attributes(mDACPrivKeyId, &attributes) == PSA_SUCCESS, CHIP_ERROR_INTERNAL);
+
+        // Get the actual block size.
+        const flash_parameters * flashParameters = flash_get_parameters(kFlashDev);
+        VerifyOrReturnError(flashParameters, CHIP_ERROR_INTERNAL);
+
+        // To write zeros directly to the Flash memory without erasing whole page the start address must be aligned to the
+        // write_block_size value (alignedDacPrivKeyOffset), then we need align the required buffer size to the write_block_size as
+        // well (requiredFlashSpaceSize) to meet Flash write requirements, and we need to calculate how many bytes of the factory
+        // data set must be copied to the additional buffer space created after the alignments (bytesToLeftBefore, and
+        // bytesToLeftAfter)
+        size_t alignedDacPrivKeyOffset = ROUND_DOWN(mFactoryData.dacPrivateKeyOffset, flashParameters->write_block_size);
+        size_t bytesToLeftBefore       = mFactoryData.dacPrivateKeyOffset % flashParameters->write_block_size;
+        size_t requiredFlashSpaceSize  = ROUND_UP(kDACPrivateKeyLength + bytesToLeftBefore, flashParameters->write_block_size);
+        size_t bytesToLeftAfter        = requiredFlashSpaceSize - bytesToLeftBefore - kDACPrivateKeyLength;
+
+        // Allocate the memory buffer for removing DAC private key.
+        chip::Platform::ScopedMemoryBuffer<uint8_t> removedPrivKeyBuffer;
+        VerifyOrReturnError(removedPrivKeyBuffer.Calloc(requiredFlashSpaceSize), CHIP_ERROR_NO_MEMORY);
+
+        // Copy the existing parts of the aligned memory space to before and after the DAC private key space.
+        memcpy(removedPrivKeyBuffer.Get(), factoryDataPartition + alignedDacPrivKeyOffset, bytesToLeftBefore);
+        memcpy(removedPrivKeyBuffer.Get() + bytesToLeftBefore + kDACPrivateKeyLength,
+               factoryDataPartition + mFactoryData.dacPrivateKeyOffset + kDACPrivateKeyLength, bytesToLeftAfter);
+
+        // Write aligned buffer directly to the Flash without erasing.
+        VerifyOrReturnError(0 ==
+                                flash_write(kFlashDev, kFactoryDataPartitionAddress + alignedDacPrivKeyOffset,
+                                            removedPrivKeyBuffer.Get(), requiredFlashSpaceSize),
+                            CHIP_ERROR_INTERNAL);
+
+        // Parse the factory data again and verify if the procedure finished successfully
+        VerifyOrReturnError(ParseFactoryData(factoryDataPartition, static_cast<uint16_t>(factoryDataSize), &mFactoryData),
+                            CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND);
+
+        // Verify if the factory data does not contain the DAC private key anymore.
+        VerifyOrReturnError(memcmp(mFactoryData.dac_priv_key.data, clearedDACPrivKey, kDACPrivateKeyLength) == 0,
+                            CHIP_ERROR_INTERNAL);
+#endif
+    }
+
     return CHIP_NO_ERROR;
 }
+#endif
 
 template <class FlashFactoryData>
 CHIP_ERROR FactoryDataProvider<FlashFactoryData>::GetCertificationDeclaration(MutableByteSpan & outBuffer)
@@ -160,8 +268,18 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::SignWithDeviceAttestationKey(c
 
     VerifyOrReturnError(outSignBuffer.size() >= signature.Capacity(), CHIP_ERROR_BUFFER_TOO_SMALL);
     ReturnErrorCodeIf(!mFactoryData.dac_cert.data, CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND);
-    ReturnErrorCodeIf(!mFactoryData.dac_priv_key.data, CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND);
 
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+    size_t outputLen = 0;
+
+    psa_status_t err = psa_sign_message(mDACPrivKeyId, PSA_ALG_ECDSA(PSA_ALG_SHA_256), messageToSign.data(), messageToSign.size(),
+                                        signature.Bytes(), signature.Capacity(), &outputLen);
+
+    VerifyOrReturnError(!err, CHIP_ERROR_INTERNAL);
+    VerifyOrReturnError(outputLen == chip::Crypto::kP256_ECDSA_Signature_Length_Raw, CHIP_ERROR_INTERNAL);
+    ReturnErrorOnFailure(signature.SetLength(outputLen));
+#else
+    ReturnErrorCodeIf(!mFactoryData.dac_priv_key.data, CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND);
     // Extract public key from DAC cert.
     ByteSpan dacCertSpan{ reinterpret_cast<uint8_t *>(mFactoryData.dac_cert.data), mFactoryData.dac_cert.len };
     chip::Crypto::P256PublicKey dacPublicKey;
@@ -171,6 +289,7 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::SignWithDeviceAttestationKey(c
         LoadKeypairFromRaw(ByteSpan(reinterpret_cast<uint8_t *>(mFactoryData.dac_priv_key.data), mFactoryData.dac_priv_key.len),
                            ByteSpan(dacPublicKey.Bytes(), dacPublicKey.Length()), keypair));
     ReturnErrorOnFailure(keypair.ECDSA_sign_msg(messageToSign.data(), messageToSign.size(), signature));
+#endif
 
     return CopySpanToMutableSpan(ByteSpan{ signature.ConstBytes(), signature.Length() }, outSignBuffer);
 }
--- a/src/platform/nrfconnect/FactoryDataProvider.h
+++ b/src/platform/nrfconnect/FactoryDataProvider.h
@@ -21,6 +21,10 @@
 #include <platform/CommissionableDataProvider.h>
 #include <platform/DeviceInstanceInfoProvider.h>
 
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+#include <crypto/CHIPCryptoPALPSA.h>
+#endif
+
 #include <fprotect.h>
 #include <pm_config.h>
 #include <system/SystemError.h>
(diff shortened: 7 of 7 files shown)


config/nrfconnect/chip-module/Kconfig
```diff
@@ -249,7 +245,7 @@ endif # CHIP_FACTORY_DATA_BUILD
 # See config/zephyr/Kconfig for full definition
 config CHIP_FACTORY_RESET_ERASE_NVS
 	bool
-	default y
+	default y if !CHIP_CRYPTO_PSA # For now erasing whole NVS sector is not supported for PSA Crypto
```
kkasperczyk-no, line 248: If this is not supported and cannot be enabled by the user, probably we should add `depends on !CHIP_CRYPTO_PSA` to the `CHIP_FACTORY_RESET_ERASE_NVS` declaration (not default override).

kkasperczyk-no: Btw. why we even redefine this Kconfig here? If we want to set the default to y we should not use the type and move it to Kconfig.defaults. Any thoughts?

config/nrfconnect/chip-module/Kconfig
```diff
@@ -284,4 +280,11 @@ config CHIP_ENABLE_READ_CLIENT
 	  This config can be disabled for device types that do not require Read Client functionality.
 	  Disabling this config can save flash and RAM space.
 
+config CHIP_CRYPTO_PSA_ITS_NVM_OFFSET
+	hex "Specify the offset for Matter crypto keys in the ITS NVM space"
+	default 0x30000
```
kkasperczyk-no, line 285: Is it some constant or is it obvious how to get this offset? For me it would be nice to add at least one sentence of explanation in the help section on how to get it from some macro or something, as it looks like a magic number.

src/platform/nrfconnect/CHIPPlatformConfig.h
```diff
@@ -54,6 +54,12 @@
 #endif // CHIP_CONFIG_SHA256_CONTEXT_ALIGN
 #endif // CONFIG_CHIP_CRYPTO_PSA
 
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+#ifndef CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE
+#define CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE CONFIG_CHIP_CRYPTO_PSA_ITS_NVM_OFFSET
+#endif
+#endif
```
kkasperczyk-no, line 61:
```suggestion
#endif // CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE
#endif // CONFIG_CHIP_CRYPTO_PSA
```

src/platform/nrfconnect/FactoryDataParser.c
```diff
@@ -84,6 +104,87 @@ static bool DecodeEntry(zcbor_state_t * states, void * buffer, size_t bufferSize
     return res;
 }
 
+bool SerializeFactoryData(uint8_t * factoryDataBuff, size_t factoryDataSize, const struct FactoryData * factoryData)
+{
+    zcbor_state_t cbor_state[2];
```
kkasperczyk-no, line 109: Why 2? Some comment?

src/platform/nrfconnect/FactoryDataParser.c
```diff
@@ -84,6 +104,87 @@ static bool DecodeEntry(zcbor_state_t * states, void * buffer, size_t bufferSize
...
+    zcbor_new_encode_state(cbor_state, MAX_FACTORY_DATA_NESTING_LEVEL, factoryDataBuff, factoryDataSize, 0);
+
+    bool res = zcbor_map_start_encode(cbor_state, 1);
+
+    res = res && zcbor_tstr_put_lit_cast(cbor_state, "version");
```
kkasperczyk-no, line 114: The result of this method will be true only if all of the below operations succeed, right? I think it would be reasonable to return after the first failure, as executing dozen of the following operations is pointless.

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -24,7 +24,15 @@
...
 
-#include <zephyr/logging/log.h>
+#include <lib/support/logging/CHIPLogging.h>
+
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+#include <platform/Zephyr/SysHeapMalloc.h>
+#include <psa/crypto.h>
+#include <zephyr/drivers/flash.h>
+
+static const struct device * const flash_dev = DEVICE_DT_GET_OR_NULL(DT_CHOSEN(zephyr_flash_controller));
```
kkasperczyk-no, line 34: Change to `kFlashDev` to keep convention?

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -95,9 +89,90 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+    // Protection against write for external storage is not supported.
+    if (error == CHIP_ERROR_NOT_IMPLEMENTED)
+    {
+        ChipLogProgress(DeviceLayer, "The device does not support hardware protection against write.");
+        error = CHIP_NO_ERROR;
+    }
+    else if (error != CHIP_NO_ERROR)
+    {
+        ChipLogError(DeviceLayer, "Failed to protect the factory data partition.");
+        return error;
```
kkasperczyk-no, line 107: Minor: I think this line could be removed and the `return CHIP_NO_ERROR` at the bottom of this method could be changed to `return error`.

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -95,9 +89,90 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+    }
+
     return CHIP_NO_ERROR;
 }
 
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+template <class FlashFactoryData>
+CHIP_ERROR FactoryDataProvider<FlashFactoryData>::MoveDACPrivateKeyToSecureStorage(uint8_t * factoryDataPartition,
```
kkasperczyk-no, line 115: I can see that `ParseFactoryData` validates the `factoryDataPartition`, so it will not crash, but maybe we could verify it at the top of this method and return prematurely to not execute multiple operations unnecessarily.

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -95,9 +89,90 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+        {
+            ChipLogError(DeviceLayer, "Could not allocate memory for removing DAC private key!");
+            return CHIP_ERROR_NO_MEMORY;
+        }
+
+        VerifyOrExit(SerializeFactoryData(reinterpret_cast<uint8_t *>(factoryDataBuff), mFactoryData.actualSize, &mFactoryData),
+                     err = CHIP_ERROR_INTERNAL);
+
+        // Replace the old factory data set with the new one.
+        VerifyOrExit(0 == flash_erase(flash_dev, PM_FACTORY_DATA_ADDRESS, factoryDataSize), err = CHIP_ERROR_NO_MEMORY);
```
kkasperczyk-no, line 163: You could use `kFactoryDataPartitionAddress` instead of `PM_FACTORY_DATA_ADDRESS`.

src/platform/nrfconnect/FactoryDataProvider.h
```diff
@@ -175,6 +183,9 @@ class FactoryDataProvider : public chip::Credentials::DeviceAttestationCredentia
 
     struct FactoryData mFactoryData;
     FlashFactoryData mFlashFactoryData;
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+    psa_key_id_t mDACPrivKeyId = static_cast<psa_key_id_t>(static_cast<uint32_t>(chip::Crypto::KeyIdBase::Maximum) + 1);
```
Damian-Nordic, line 187: I think you should rather reserve a value in `KeyIdBase` than assuming `max + 1` is not reserved and can be used for DAC.

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -61,6 +61,12 @@ config CHIP_MALLOC_SYS_HEAP_SIZE
 config NVS_LOOKUP_CACHE_SIZE
     default 512
 
+# See config/zephyr/Kconfig for full definition
+config CHIP_FACTORY_RESET_ERASE_NVS
```
Damian-Nordic, line 65: Why is this moved? It could be kept in the previous place even if the condition on `!CHIP_CRYPTO_PSA` was needed.

config/nrfconnect/chip-module/Kconfig
Damian-Nordic: Not sure if this is useful to have this exposed to Kconfig (it's probably us who need to make sure PSA key ID space is divided properly across components and customers usually won't be interested in changing that), but if you think otherwise please, remove "Specify the" part from the prompt to keep the style more consistent with others.

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -95,9 +89,95 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+    // Check if factory data contains DAC private key
+    if (mFactoryData.dacPrivateKeyPresent)
+    {
+        psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+        // If there is the new DAC private key present in the factory data set and also there is
+        // the existing one in ITS NVM storage, then it is a security violation
+        // TODO: Consider other reactions to this violation: blocking the application, factory reset, etc.
+        if (psa_get_key_attributes(mDACPrivKeyId, &attributes) != PSA_SUCCESS)
+        {
+            ChipLogDetail(DeviceLayer, "Found DAC Private Key in factory data set. Copying to ITS...");
```
Damian-Nordic, line 136:
```suggestion
            ChipLogProgress(DeviceLayer, "Found DAC Private Key in factory data set. Moving to ITS...");
```

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -95,9 +89,95 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+                                        kDACPrivateKeyLength, &mDACPrivKeyId) == PSA_SUCCESS,
+                         err = CHIP_ERROR_INTERNAL);
+        }
+
+        memset(mFactoryData.dac_priv_key.data, 0xff, mFactoryData.dac_priv_key.len);
+        mFactoryData.dacPrivateKeyPresent = false;
+
+        // Allocate needed memory space to perform removal and moving DAC private key
+        size_t alignedSize = mFactoryData.actualSize + (sizeof(uint32_t) - (mFactoryData.actualSize % sizeof(uint32_t)));
+        factoryDataBuff    = chip::DeviceLayer::Malloc::Calloc(alignedSize, sizeof(uint8_t));
```
Damian-Nordic, line 156: You shouldn't assume sys_heap malloc is enabled. Please use `Platform::Calloc` or (even better) `ScopedMemoryBuffer<uint8_t>` to automatically release the pointer.

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -95,9 +89,95 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+
+        if (!factoryDataBuff)
+        {
+            ChipLogError(DeviceLayer, "Could not allocate memory for removing DAC private key!");
+            return CHIP_ERROR_NO_MEMORY;
+        }
+
+        VerifyOrExit(SerializeFactoryData(reinterpret_cast<uint8_t *>(factoryDataBuff), mFactoryData.actualSize, &mFactoryData),
```
Damian-Nordic, line 164: Why do we need to serialize the factory data? Wouldn't it be much easier to simply set the DAC private key to ffff... and save the original CBOR with just this small modification? Then we wouldn't need to pull ZCBOR encoding code and it should be less error-prone (we don't have to keep the parsing and serializing code in sync in the future).

config/nrfconnect/chip-module/Kconfig.defaults
Damian-Nordic: Oh, Kamil asked for that :). So it was overriden in `nrfconnect/.../Kconfig` only because Telink had a different default but this setting was still part of the Matter Zephyr module, so it seemed OK to put define it there instead of in `Kconfig.defaults` but whatever :).

src/platform/nrfconnect/FactoryDataParser.c
```diff
@@ -84,6 +104,105 @@ static bool DecodeEntry(zcbor_state_t * states, void * buffer, size_t bufferSize
     return res;
 }
 
+bool SerializeFactoryData(uint8_t * factoryDataBuff, size_t factoryDataSize, const struct FactoryData * factoryData)
+{
+    zcbor_state_t cbor_state[MAX_FACTORY_DATA_NESTING_LEVEL];
+    zcbor_new_encode_state(cbor_state, MAX_FACTORY_DATA_NESTING_LEVEL, factoryDataBuff, factoryDataSize, 0);
+
+    if (!(zcbor_map_start_encode(cbor_state, 1) && zcbor_tstr_put_lit_cast(cbor_state, "version") &&
+          zcbor_uint32_put(cbor_state, (uint32_t) (factoryData->version)) && zcbor_tstr_put_lit_cast(cbor_state, "passcode") &&
```
markaj-nordic, line 113: Can we keep C++ style for casting? (static_cast would fit)

src/platform/nrfconnect/FactoryDataParser.c
```diff
@@ -84,6 +104,105 @@ static bool DecodeEntry(zcbor_state_t * states, void * buffer, size_t bufferSize
     return res;
 }
 
+bool SerializeFactoryData(uint8_t * factoryDataBuff, size_t factoryDataSize, const struct FactoryData * factoryData)
+{
+    zcbor_state_t cbor_state[MAX_FACTORY_DATA_NESTING_LEVEL];
+    zcbor_new_encode_state(cbor_state, MAX_FACTORY_DATA_NESTING_LEVEL, factoryDataBuff, factoryDataSize, 0);
+
+    if (!(zcbor_map_start_encode(cbor_state, 1) && zcbor_tstr_put_lit_cast(cbor_state, "version") &&
```
markaj-nordic, line 112: I'd think of a private helper function like DoSerialize which would wrap lines 112-142 and here only call `VerifyOrReturnValue(DoSerialize(), false)` or similar.

src/platform/nrfconnect/FactoryDataProvider.cpp
markaj-nordic: +1 for `ScopedMemoryBuffer` or similar automatic memory holder primitive. Also we don't need to define `factoryDataBuff` at the beginning of the function, it can be instantiated directly when assigning the pointer here.

src/platform/nrfconnect/FactoryDataParser.c
markaj-nordic: edit: valid only if we keep the serialization instead of using factory data with invalidated DAC.

src/platform/nrfconnect/FactoryDataProvider.h
ArekBalysNordic: Ok, I was considering how to do it, and reserving it in KeyIdBase seems good.

src/platform/nrfconnect/FactoryDataParser.c
ArekBalysNordic: It's a C file.

src/platform/nrfconnect/FactoryDataParser.c
ArekBalysNordic: Serialization will be removed.

config/nrfconnect/chip-module/Kconfig.defaults
ArekBalysNordic: Maybe let's keep it in Kconfig as it was, but we need to add the dependency to PSA Crypto until we have a better way to keep keys alive after performing the factory reset.

src/platform/nrfconnect/FactoryDataProvider.cpp
ArekBalysNordic: Right, I thought that we may need it in the future, but after consideration, I agree that we won't need to modify the factory data in any other way than removing DAC. So I will simply overwrite the dac priv key and flash the new factory data set.

ArekBalysNordic:
@Damian-Nordic @markaj-nordic @kkasperczyk-no  I've removed serialization from the factory data parser and overwritten the dac private key directly from the factory data set. Please look at it once again :)

src/platform/nrfconnect/FactoryDataParser.c
```diff
@@ -198,70 +234,88 @@ bool ParseFactoryData(uint8_t * buffer, uint16_t bufferSize, struct FactoryData
...
             res                              = res && uint8_decode(states, &factoryData->primary_color);
             factoryData->primaryColorPresent = res;
+            updateSize(factoryData, "primary_color", sizeof(uint32_t));
         }
         else if (strncmp("user", (const char *) currentString.value, currentString.len) == 0)
         {
             factoryData->user.data = (void *) states->payload;
             res                    = res && zcbor_any_skip(states, NULL);
             factoryData->user.len  = (size_t) ((void *) states->payload - factoryData->user.data);
+            updateSize(factoryData, "user", factoryData->user.len);
```
Damian-Nordic, line 317: I guess you could use `states->payload - buffer` after `zcbor_list_map_end_force_decode` to learn the total CBOR length, right? Then all these `updateSize` calls wouldn't be needed.

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -24,7 +24,15 @@
...
 
-#include <zephyr/logging/log.h>
+#include <lib/support/logging/CHIPLogging.h>
+
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+#include <lib/support/ScopedBuffer.h>
+#include <psa/crypto.h>
+#include <zephyr/drivers/flash.h>
+
+static const struct device * const kFlashDev = DEVICE_DT_GET_OR_NULL(DT_CHOSEN(zephyr_flash_controller));
```
Damian-Nordic, line 34: This seems to assume the factory data in external flash is not supported. As more and more factory data features are not supported using external flash I think it arguable if we need the external flash support at all in the current state (just saying, not asking for any changes in this PR :)).

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -95,8 +89,95 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+    return error;
+}
+
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+template <class FlashFactoryData>
+CHIP_ERROR FactoryDataProvider<FlashFactoryData>::MoveDACPrivateKeyToSecureStorage(uint8_t * factoryDataPartition,
+                                                                                   size_t factoryDataSize)
+{
+
+    CHIP_ERROR err = CHIP_NO_ERROR;
```
Damian-Nordic, line 119: Unused

src/platform/nrfconnect/FactoryDataParser.c
```diff
@@ -198,70 +234,88 @@ bool ParseFactoryData(uint8_t * buffer, uint16_t bufferSize, struct FactoryData
...
         else if (strncmp("dac_cert", (const char *) currentString.value, currentString.len) == 0)
         {
             res = res && zcbor_bstr_decode(states, (struct zcbor_string *) &factoryData->dac_cert);
+            updateSize(factoryData, "dac_cert", factoryData->dac_cert.len);
         }
         else if (strncmp("dac_key", (const char *) currentString.value, currentString.len) == 0)
         {
             res = res && zcbor_bstr_decode(states, (struct zcbor_string *) &factoryData->dac_priv_key);
+            updateSize(factoryData, "dac_key", factoryData->dac_priv_key.len);
+            factoryData->dac_priv_key_offset = (size_t) ((void *) states->payload - (void *) buffer) - factoryData->dac_priv_key.len;
```
Damian-Nordic, line 253:
```suggestion
            factoryData->dac_priv_key_offset = (uint8_t*)factoryData->dac_priv_key.data - buffer;
```

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -95,8 +89,95 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+        {
+            ChipLogProgress(DeviceLayer, "Found DAC Private Key in factory data set. Copying to ITS...");
+
+            psa_reset_key_attributes(&attributes);
+            psa_set_key_type(&attributes, PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_SECP_R1));
+            psa_set_key_bits(&attributes, kDACPrivateKeyLength * 8);
+            psa_set_key_algorithm(&attributes, PSA_ALG_ECDSA(PSA_ALG_SHA_256));
+            psa_set_key_lifetime(&attributes, PSA_KEY_LIFETIME_PERSISTENT);
+            psa_set_key_id(&attributes, mDACPrivKeyId);
+            psa_set_key_usage_flags(&attributes, PSA_KEY_USAGE_EXPORT | PSA_KEY_USAGE_SIGN_MESSAGE);
```
Damian-Nordic, line 145: EXPORT shouldn't be needed

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -95,8 +89,95 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+    uint8_t ClearedDACPrivKey[kDACPrivateKeyLength];
+    memset(ClearedDACPrivKey, 0xFF, sizeof(ClearedDACPrivKey));
+
+    if (!factoryDataPartition || factoryDataSize == 0)
+    {
+        return CHIP_ERROR_INVALID_ARGUMENT;
+    }
+
+    // Check if factory data contains DAC private key
+    if (memcmp(mFactoryData.dac_priv_key.data, ClearedDACPrivKey, kDACPrivateKeyLength) != 0)
```
Damian-Nordic, line 129: The parser should verify that the actual key length is kDACPrivateKeyLength before this is executed.

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -95,8 +89,95 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+            psa_set_key_id(&attributes, mDACPrivKeyId);
+            psa_set_key_usage_flags(&attributes, PSA_KEY_USAGE_EXPORT | PSA_KEY_USAGE_SIGN_MESSAGE);
+
+            VerifyOrReturnError(psa_import_key(&attributes, reinterpret_cast<uint8_t *>(mFactoryData.dac_priv_key.data),
+                                               kDACPrivateKeyLength, &mDACPrivKeyId) == PSA_SUCCESS,
+                                CHIP_ERROR_INTERNAL);
+        }
+
+        // Allocate needed memory space to perform removal and moving DAC private key
+        size_t alignedSize = mFactoryData.actualSize + (sizeof(uint32_t) - (mFactoryData.actualSize % sizeof(uint32_t)));
```
Damian-Nordic, line 153: Curiosity: why do we need to align the size? If necessary, you can use Zephyr's `ROUND_UP` macro, as the current formula will read 4 bytes more than necessary if the size is already rounded. Also, if this is because of `flash_write` requirements, you could use `flash_get_parameters` to obtain real `write_block_size` to make the solution ready for future platforms.

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -95,8 +89,95 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+
+        // Copy existing factoryDataSet
+        memcpy(factoryDataBuff.Get(), factoryDataPartition, alignedSize);
+
+        // Overwrite the DAC private key
+        memcpy(factoryDataBuff.Get() + mFactoryData.dac_priv_key_offset, ClearedDACPrivKey, kDACPrivateKeyLength);
+
+        // Replace the old factory data set with the new one.
+        VerifyOrReturnError(0 == flash_erase(kFlashDev, kFactoryDataPartitionAddress, kFactoryDataPartitionSize),
+                            CHIP_ERROR_NO_MEMORY);
```
Damian-Nordic, line 165: `CHIP_ERROR_NO_MEMORY` is used for failed memory allocation, so another code should be used.

Damian-Nordic (commented):
Found a few more things to remove or simplify.

src/platform/nrfconnect/FactoryDataParser.c
ArekBalysNordic: Oh, right :)

src/platform/nrfconnect/FactoryDataProvider.cpp
ArekBalysNordic: Yeah, I completely agree with that, I think we shouldn't suggest using factory data in an external flash at all :)

src/platform/nrfconnect/FactoryDataProvider.cpp
ArekBalysNordic: Yes, it is because of `flash_write`. Using `write_block_size` and aligning to that value sounds good.

src/platform/nrfconnect/FactoryDataParser.c
```diff
@@ -17,12 +17,32 @@
...
 #include <zcbor_decode.h>
+#include <zcbor_encode.h>
 
 #include <ctype.h>
 #include <string.h>
 
-#define MAX_FACTORY_DATA_NESTING_LEVEL 4
+#define zcbor_tstr_put_lit_cast(state, string) zcbor_tstr_encode_ptr(state, (char *) string, sizeof(string) - 1)
+
+#define MAX_FACTORY_DATA_NESTING_LEVEL 2
```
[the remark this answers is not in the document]
ArekBalysNordic: I can see that we previously reduced the nesting level because we don't parse the `user_data` field using the factory data parser, but we didn't change that value. We have only two levels in the factory data set, a map and then values.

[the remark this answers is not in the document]
Damian-Nordic: I see, thanks

Damian-Nordic (approved):

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -73,30 +107,83 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+                            CHIP_ERROR_INTERNAL);
+        VerifyOrReturnError(0 == flash_write(kFlashDev, kFactoryDataPartitionAddress, factoryDataBuff.Get(), alignedSize),
+                            CHIP_ERROR_INTERNAL);
+
+        // Parse the factory Data again and verify if the procedure finished successfully
+        VerifyOrReturnError(ParseFactoryData(factoryDataPartition, static_cast<uint16_t>(factoryDataSize), &mFactoryData),
+                            CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND);
+
+        // Verify if the factory data does not contain the DAC private key anymore.
+        VerifyOrReturnError(memcmp(mFactoryData.dac_priv_key.data, ClearedDACPrivKey, kDACPrivateKeyLength) == 0,
```
markaj-nordic, line 180: I wonder if we should consider some cleanup when this check fails. Is it possible that only some parts of the DAC will be stored? Wouldn't it then pose some issues when detecting if the DAC was present at the next boot? (we do die if the `MoveDACPrivateKeyToSecureStorage` returns error)

markaj-nordic (approved):

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -73,30 +107,83 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
-        return error;
+        return CHIP_ERROR_INVALID_ARGUMENT;
     }
 
-    if (!ParseFactoryData(factoryData, static_cast<uint16_t>(factoryDataSize), &mFactoryData))
+    if (mFactoryData.dac_priv_key.len != kDACPrivateKeyLength)
     {
-        ChipLogError(DeviceLayer, "Failed to parse factory data");
-        return CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND;
+        return CHIP_ERROR_INCORRECT_STATE;
```
kkasperczyk-no, line 126: Maybe it would be better to return some error related to invalid length instead of incorrect state?

src/platform/nrfconnect/FactoryDataParser.h
```diff
@@ -66,6 +66,8 @@ struct FactoryData
     bool discriminatorPresent;
     bool productFinishPresent;
     bool primaryColorPresent;
+    size_t actualSize;
+    size_t dac_priv_key_offset;
```
kkasperczyk-no, line 70: Please align case to the other struct members

kkasperczyk-no (approved):

src/platform/nrfconnect/FactoryDataProvider.cpp
ArekBalysNordic: Yeah, this seems to be a problem, especially for the existing devices, so we need to simplify the process as much as possible and also create a possibility of not removing the DAC from the FD. I've added an additional kconfig to `CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY` to enable or disable removing DAC from the FD, so users can decide if they take a risk. Moreover, the flash driver allows writing zeros without erasing the whole page (but only twice without erasing), and in this way, we can remove only a DAC key without reflashing the whole FD set - it reduces the chance of failure. We will need to describe it properly in the migration guide - as a followup.

ArekBalysNordic:
@Damian-Nordic @LuDuda I've provided the changes that were discussed F2F, now we have the config CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY to decide if we want to clear DAC or not, and I've changed a way of removing dac. Could you please verify the recent changes?

config/nrfconnect/chip-module/Kconfig
```diff
@@ -284,4 +282,16 @@ config CHIP_ENABLE_READ_CLIENT
 	  This config can be disabled for device types that do not require Read Client functionality.
 	  Disabling this config can save flash and RAM space.
 
+config CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
+	bool "Migrate DAC private key from factory data to PSA ITS"
+	default y
```
Damian-Nordic, line 287: This should be `y` only if ITS is encrypted (either TF-M or ITS with HUK is used - not sure what's Kconfig for the latter).

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -73,30 +107,93 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+            psa_reset_key_attributes(&attributes);
+            psa_set_key_type(&attributes, PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_SECP_R1));
+            psa_set_key_bits(&attributes, kDACPrivateKeyLength * 8);
+            psa_set_key_algorithm(&attributes, PSA_ALG_ECDSA(PSA_ALG_SHA_256));
+#ifdef CONFIG_CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
+            psa_set_key_lifetime(&attributes, PSA_KEY_LIFETIME_PERSISTENT);
+#else
+            psa_set_key_lifetime(&attributes, PSA_KEY_LIFETIME_VOLATILE);
+#endif
+            psa_set_key_id(&attributes, mDACPrivKeyId);
```
Damian-Nordic, line 151: This is also only for `CONFIG_CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY`

config/nrfconnect/chip-module/Kconfig
```diff
@@ -250,6 +246,8 @@ endif # CHIP_FACTORY_DATA_BUILD
 config CHIP_FACTORY_RESET_ERASE_NVS
 	bool
 	default y
+	depends on !CHIP_CRYPTO_PSA
```
Damian-Nordic, line 249: You can use it with PSA as long as you don't migrate DAC key so this is not needed.

src/platform/nrfconnect/FactoryDataParser.c
```diff
@@ -269,5 +277,8 @@ bool ParseFactoryData(uint8_t * buffer, uint16_t bufferSize, struct FactoryData
         }
     }
 
-    return res && zcbor_list_map_end_force_decode(states);
+    res                     = res && zcbor_list_map_end_force_decode(states);
+    factoryData->actualSize = (size_t) (states->payload - buffer);
```
Damian-Nordic, line 281: Not needed anymore

src/platform/nrfconnect/FactoryDataParser.c
```diff
@@ -17,12 +17,14 @@
 
 #include "FactoryDataParser.h"
 
+#include <zcbor_common.h>
 #include <zcbor_decode.h>
+#include <zcbor_encode.h>
```
Damian-Nordic, line 22: Not needed anymore.

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -73,30 +107,93 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+        size_t alignedDacPrivKeyOffset = ROUND_DOWN(mFactoryData.dacPrivateKeyOffset, flashParameters->write_block_size);
+        size_t bytesToLeftBefore       = mFactoryData.dacPrivateKeyOffset % flashParameters->write_block_size;
+        size_t requiredFlashSpaceSize  = ROUND_UP(kDACPrivateKeyLength + bytesToLeftBefore, flashParameters->write_block_size);
+        size_t bytesToLeftAfter        = requiredFlashSpaceSize - bytesToLeftBefore - kDACPrivateKeyLength;
+
+        // Allocate the memory buffer for removing DAC private key.
+        chip::Platform::ScopedMemoryBuffer<uint8_t> removedPrivKeyBuffer;
+        VerifyOrReturnError(removedPrivKeyBuffer.Calloc(requiredFlashSpaceSize), CHIP_ERROR_NO_MEMORY);
+        // Copy the existing parts of the aligned memory space to before and after the DAC private key space.
+        memcpy(removedPrivKeyBuffer.Get(), factoryDataPartition, bytesToLeftBefore);
```
Damian-Nordic, line 178: Source pointers seem incorrect.

config/nrfconnect/chip-module/Kconfig
ArekBalysNordic: Ok, so `CONFIG_BUILD_WITH_TFM` or `CONFIG_SECURE_STORAGE_BACKEND_AEAD_KEY_DERIVE_FROM_HUK`

Damian-Nordic (approved):

src/crypto/CHIPCryptoPALPSA.h
```diff
@@ -62,7 +62,8 @@ enum class KeyIdBase : psa_key_id_t
 {
     Minimum     = CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE,
     Operational = Minimum, ///< Base of the PSA key ID range for Node Operational Certificate private keys
-    Maximum     = Operational + kMaxValidFabricIndex,
+    DACPrivKey  = Operational + kMaxValidFabricIndex + 1,
```
LuDuda, line 65: Isn't that risky, what if someone decide to increase number of  fabrics in the future with DFU? Should we keep DAC in first well-known position?

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -73,30 +107,94 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+        return CHIP_ERROR_INVALID_LIST_LENGTH;
     }
 
-    // Check if factory data version is correct
-    if (mFactoryData.version != CONFIG_CHIP_FACTORY_DATA_VERSION)
+    uint8_t clearedDACPrivKey[kDACPrivateKeyLength];
+    memset(clearedDACPrivKey, 0x00, sizeof(clearedDACPrivKey));
+
+    // Check if factory data contains DAC private key
+    if (memcmp(mFactoryData.dac_priv_key.data, clearedDACPrivKey, kDACPrivateKeyLength) != 0)
```
LuDuda, line 133: I think we discussed that first we should check the existence of PSA key, not the content of factory data, right?

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -73,30 +107,94 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::Init()
...
+        // bytesToLeftAfter)
+        size_t alignedDacPrivKeyOffset = ROUND_DOWN(mFactoryData.dacPrivateKeyOffset, flashParameters->write_block_size);
+        size_t bytesToLeftBefore       = mFactoryData.dacPrivateKeyOffset % flashParameters->write_block_size;
+        size_t requiredFlashSpaceSize  = ROUND_UP(kDACPrivateKeyLength + bytesToLeftBefore, flashParameters->write_block_size);
+        size_t bytesToLeftAfter        = requiredFlashSpaceSize - bytesToLeftBefore - kDACPrivateKeyLength;
+
+        // Allocate the memory buffer for removing DAC private key.
+        chip::Platform::ScopedMemoryBuffer<uint8_t> removedPrivKeyBuffer;
+        VerifyOrReturnError(removedPrivKeyBuffer.Calloc(requiredFlashSpaceSize), CHIP_ERROR_NO_MEMORY);
+        // Copy the existing parts of the aligned memory space to before and after the DAC private key space.
```
LuDuda, line 177: super nit: for visibility, can we make a new line before "//" comments?

LuDuda (changes requested):

src/crypto/CHIPCryptoPALPSA.h
LuDuda: Ahh.. it's the UINT8_MAX not the current max fabric.. Then it probably all is okay.

ArekBalysNordic: So will be ok if the DACPrivKey is set to the Minimum value within that enum?

src/platform/nrfconnect/FactoryDataProvider.cpp
ArekBalysNordic: Maybe I missed that. I will change it then.

Damian-Nordic: Hmm. I didn't remember that. :) We still want to clear the FD even if the key is already imported, right?

ArekBalysNordic: But honestly, what is the difference? What should we do if the key already exists in PSA ITS: just exit the function, or skip importing and remove the dac key if the CONFIG_CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY is set?

src/crypto/CHIPCryptoPALPSA.h
LuDuda: I think it's all okay. I though that  `kMaaxValidFabricIndex` indicates the current max supported fabrics e.g. 5. But in fact it means 254.. so it's fine.

src/platform/nrfconnect/FactoryDataProvider.cpp
LuDuda: Hmm.. I thought we agreed on sth like:

```
If PSA key does not exist:
   Import Key from factory data

#if CONFIG_CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY 
if DAC is not empty:
   Clear the DAC in factory data
#endif
```

But indeed.. now i don't see any edge case with resetting even if we make an if-statement on DAC in factory reset. I believe we discussed one option, when DAC might be partially overwritten due to reset in between.. though i think it's not the issue..

LuDuda (approved):

PR #373, opened 2024-01-03
[nrf noup][zephyr] Enabled support for bonding in the BLEManager

kkasperczyk-no:
Currently the Zephyr BLE Manager rotates Bluetooth addresses on every boot and it does this by creating new Bluetooth identity. Because of that the Zephyr stack does not create default Bluetooth ID, which is required e.g. for the bonding purposes.

Added creating two separate Bluetooth identities - for the Matter service advertising and for the bonding purposes.

### DIFF
--- a/src/platform/Zephyr/BLEManagerImpl.cpp
+++ b/src/platform/Zephyr/BLEManagerImpl.cpp
@@ -43,6 +43,8 @@
 #include <zephyr/sys/byteorder.h>
 #include <zephyr/sys/util.h>
 
+#include <zephyr/settings/settings.h>
+
 #include <array>
 
 using namespace ::chip;
@@ -106,7 +108,13 @@ bt_gatt_service sChipoBleService = BT_GATT_SERVICE(sChipoBleAttributes);
 // This value should be adjusted accordingly if the service declaration changes.
 constexpr int kCHIPoBLE_CCC_AttributeIndex = 3;
 
-CHIP_ERROR InitRandomStaticAddress()
+#ifdef CONFIG_BT_BONDABLE
+constexpr uint8_t kMatterBleIdentity = 1;
+#else
+constexpr uint8_t kMatterBleIdentity = 0;
+#endif // CONFIG_BT_BONDABLE
+
+int InitRandomStaticAddress(bool idPresent, int & id)
 {
     // Generate a random static address for the default identity.
     // This must be done before bt_enable() as after that updating the default identity is not possible.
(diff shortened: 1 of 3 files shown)


src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -151,10 +165,33 @@ CHIP_ERROR BLEManagerImpl::_Init()
...
+
+    err = bt_enable(nullptr);
+
     VerifyOrReturnError(err == 0, MapErrorZephyr(err));
 
+    settings_load();
+
+    bt_id_get(idsAddr, &idsCount);
+
+    err = InitRandomStaticAddress(idsCount > 1 ? true : false, id);
```
Damian-Nordic, line 182:
```suggestion
    err = InitRandomStaticAddress(idsCount > 1, id);
```
comparison already returns bool

src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -121,20 +123,29 @@ CHIP_ERROR InitRandomStaticAddress()
...
     {
         ChipLogError(DeviceLayer, "Failed to create BLE identity: %d", error);
-        return System::MapErrorZephyr(error);
+        return id;
     }
 
     ChipLogProgress(DeviceLayer, "BLE address: %02X:%02X:%02X:%02X:%02X:%02X", addr.a.val[5], addr.a.val[4], addr.a.val[3],
                     addr.a.val[2], addr.a.val[1], addr.a.val[0]);
-    return CHIP_NO_ERROR;
+    return error;
```
Damian-Nordic, line 148:
```suggestion
    return 0;
```
Save readers the effort of realizing that 0 is the only possibility in this place :)

src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -151,10 +165,33 @@ CHIP_ERROR BLEManagerImpl::_Init()
 
     memset(mSubscribedConns, 0, sizeof(mSubscribedConns));
 
-    ReturnErrorOnFailure(InitRandomStaticAddress());
-    int err = bt_enable(NULL);
+#ifdef CONFIG_BT_BONDABLE
+    static_assert(CONFIG_BT_ID_MAX == 2);
```
Damian-Nordic, line 169: Is that actually necessary? Users might add more IDs and that's fine, right?

src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -151,10 +165,33 @@ CHIP_ERROR BLEManagerImpl::_Init()
...
+
     VerifyOrReturnError(err == 0, MapErrorZephyr(err));
 
+    settings_load();
+
+    bt_id_get(idsAddr, &idsCount);
+
+    err = InitRandomStaticAddress(idsCount > 1 ? true : false, id);
+
+    VerifyOrReturnError(err == 0 && id == 1, MapErrorZephyr(err));
```
Damian-Nordic, line 184: Consider some constant like `kMatterBleIdentity = 1` and use it where needed.

Damian-Nordic (approved):
Looks good! Some readability nits from me.

markaj-nordic (approved):

src/platform/Zephyr/BLEManagerImpl.cpp
kkasperczyk-no: The problem with bigger number of ids than 1 (no bonding) or 2 (bonding + Matter) is that it's hard to know which one of them should be used to what purpose after loading it from settings. For id 0 it obvious (it's default), but if user created one more beside the Matter's one, we are going to load ids 1 and 2. The order of creation will decide about the id value and I was afraid we will not be able to figure it out.

Nevertheless we create Matter id in the `Init` method, so maybe it's safe to assume it will be always equal to 1, and the user's ids will be created using following numbers.

Damian-Nordic: Yes, I think that given that you call this just after `bt_enable`, it will always be 1.

ArekBalysNordic (approved):

src/platform/Zephyr/BLEManagerImpl.cpp
kkasperczyk-no: Ok, applied.

PR #374, opened 2024-01-04
[nrf noup] Migrate Operational Keys from mbedTLS to PSA ITS

ArekBalysNordic:
Implemented a mechanism to detect all existing mbedTLS-related operational keys stored in settings (mt/f/%d/o) and move them to the PSA ITS secure storage.

After the migration all keys will be removed from settings space.


src/crypto/PSAOperationalKeystore.cpp
```diff
@@ -135,6 +144,36 @@ CHIP_ERROR PSAOperationalKeystore::NewOpKeypairForFabric(FabricIndex fabricIndex
...
+    psa_set_key_algorithm(&attributes, PSA_ALG_ECDSA(PSA_ALG_SHA_256));
+    psa_set_key_usage_flags(&attributes, PSA_KEY_USAGE_SIGN_MESSAGE);
+    psa_set_key_lifetime(&attributes, PSA_KEY_LIFETIME_PERSISTENT);
+    psa_set_key_id(&attributes, GetKeyId());
+
+    status = psa_import_key(&attributes, input.ConstBytes() + mPublicKey.Length(), kP256_PrivateKey_Length, &keyId);
+    VerifyOrExit(status == PSA_SUCCESS, error = CHIP_ERROR_INTERNAL);
+
+    bbuf.Put(input.ConstBytes(), mPublicKey.Length());
+    VerifyOrExit(bbuf.Fit(), error = CHIP_ERROR_NO_MEMORY);
```
LuDuda, line 169: What happens with key that is imported at line 166 if this error occurs?

LuDuda (approved):
What about to contributing this to upstream as well, to get more reviews?

ArekBalysNordic:
> What about to contributing this to upstream as well, to get more reviews?

Sure, why not, we can start pushing PSA changes into upstream. :)

src/crypto/PSAOperationalKeystore.cpp
ArekBalysNordic: I guess it should destroy a private key if the error occurs because we need to have a valid keypair. I will add it.

src/crypto/PSAOperationalKeystore.cpp
```diff
@@ -209,5 +254,54 @@ void PSAOperationalKeystore::ReleasePendingKeypair()
...
+            ReturnErrorOnFailure(reader.Get(opKeyVersion));
+            VerifyOrReturnError(opKeyVersion == kOpKeyVersion, CHIP_ERROR_VERSION_MISMATCH);
+            ReturnErrorOnFailure(reader.Next(kOpKeyDataTag));
+
+            ByteSpan keyData;
+            ReturnErrorOnFailure(reader.GetByteView(keyData));
+            ReturnErrorOnFailure(reader.ExitContainer(containerType));
+            ReturnErrorOnFailure(reader.VerifyEndOfContainer());
+
+            if (keyData.size() == kP256_PrivateKey_Length + kP256_PublicKey_Length)
```
Damian-Nordic, line 287: nit: I would error out if that happens

Damian-Nordic (approved):

ArekBalysNordic:
I've changed the way to migrate Operational Keys. Now the keys are migrated in the Server::Init method, just after getting info about FabricTable. If the key cannot be migrated, then the new ChipEvent is posted and the Server::Init method doesn't fail. I've added two methods to OperationalKeystore API to migrate and export keys, and implemented both virtual methods for PersistentOperationalKeystore and PSAOperationalKeystore. And also I've written tests for migration and exporting. 

Please review those changes before I will go upstream with them.

src/crypto/OperationalKeystore.h
```diff
@@ -119,6 +119,35 @@ class OperationalKeystore
...
+    /**
+     * @brief Try to read out the permanently commited operational keypair and save it to the buffer.
+     *
+     * @param fabricIndex - FabricIndex from which the keypair will be exported
+     * @param outKeypair - a reference to P256SerializedKeypair object to store the exported key.
+     * @retval CHIP_ERROR on success.
+     * @retval CHIP_ERROR_UNSUPPORTED_CHIP_FEATURE if the key cannot be exported due to security restrictions.
+     * @retval CHIP_ERROR_NOT_IMPLEMENTED if the exporting is not implemented for the crypto engine.
+     * @retval CHIP_ERROR_INVALID_FABRIC_INDEX if there is no keypair found for `fabricIndex`.
+     * @retval CHIP_ERROR_BUFFER_TOO_SMALL if `keyPair` buffer is too small to store the read out keypair.
```
Damian-Nordic, line 131:
```suggestion
     * @retval CHIP_ERROR_BUFFER_TOO_SMALL if `outKeyPair` buffer is too small to store the read out keypair.
```

src/crypto/PSAOperationalKeystore.cpp
```diff
@@ -154,6 +184,33 @@ CHIP_ERROR PSAOperationalKeystore::CommitOpKeypairForFabric(FabricIndex fabricIn
     return CHIP_NO_ERROR;
 }
 
+CHIP_ERROR PSAOperationalKeystore::ExportOpKeypairForFabric(FabricIndex fabricIndex, Crypto::P256SerializedKeypair & outKeypair)
+{
+    VerifyOrReturnError(IsValidFabricIndex(fabricIndex), CHIP_ERROR_INVALID_FABRIC_INDEX);
```
Damian-Nordic, line 189: This implementation imports keys as non-exportable, so this method has no chance to succeed. You can just return _not supported_ error.

src/crypto/PSAOperationalKeystore.cpp
```diff
@@ -209,5 +266,32 @@ void PSAOperationalKeystore::ReleasePendingKeypair()
     mIsPendingKeypairActive = false;
 }
 
+CHIP_ERROR PSAOperationalKeystore::MigrateOpKeypairForFabric(FabricIndex fabricIndex,
+                                                             OperationalKeystore & operationalKeystore) const
+{
+    VerifyOrReturnError(IsValidFabricIndex(fabricIndex), CHIP_ERROR_INVALID_FABRIC_INDEX);
+    VerifyOrReturnError(fabricIndex != kUndefinedFabricIndex, CHIP_ERROR_INVALID_FABRIC_INDEX);
```
Damian-Nordic, line 273: You just checked that the fabric index is valid so this is redundant.

src/crypto/PersistentStorageOperationalKeystore.cpp
```diff
@@ -310,4 +325,38 @@ void PersistentStorageOperationalKeystore::ReleaseEphemeralKeypair(Crypto::P256K
     Platform::Delete<Crypto::P256Keypair>(keypair);
 }
 
+CHIP_ERROR PersistentStorageOperationalKeystore::MigrateOpKeypairForFabric(FabricIndex fabricIndex,
+                                                                           OperationalKeystore & operationalKeystore) const
+{
+    VerifyOrReturnError(mStorage, CHIP_ERROR_INCORRECT_STATE);
+    VerifyOrReturnError(IsValidFabricIndex(fabricIndex), CHIP_ERROR_INVALID_FABRIC_INDEX);
+    VerifyOrReturnError(fabricIndex != kUndefinedFabricIndex, CHIP_ERROR_INVALID_FABRIC_INDEX);
```
Damian-Nordic, line 333: Redundant

src/crypto/PersistentStorageOperationalKeystore.cpp
```diff
@@ -310,4 +325,38 @@ void PersistentStorageOperationalKeystore::ReleaseEphemeralKeypair(Crypto::P256K
     Platform::Delete<Crypto::P256Keypair>(keypair);
 }
 
+CHIP_ERROR PersistentStorageOperationalKeystore::MigrateOpKeypairForFabric(FabricIndex fabricIndex,
+                                                                           OperationalKeystore & operationalKeystore) const
+{
+    VerifyOrReturnError(mStorage, CHIP_ERROR_INCORRECT_STATE);
```
Damian-Nordic, line 331:
```suggestion
    VerifyOrReturnError(mStorage != nullptr, CHIP_ERROR_INCORRECT_STATE);
```
nit: but in the upstream devs often ask for full condition to have the code more self-explanatory.

src/crypto/PSAOperationalKeystore.cpp
```diff
@@ -209,5 +266,32 @@ void PSAOperationalKeystore::ReleasePendingKeypair()
...
+        if (!HasOpKeypairForFabric(fabricIndex))
+        {
+            PersistentP256Keypair keypair(fabricIndex);
+            ReturnErrorOnFailure(keypair.Deserialize(serializedKeypair));
+        }
+
+        ReturnErrorOnFailure(operationalKeystore.RemoveOpKeypairForFabric(fabricIndex));
+    }
+
+    return err;
```
Damian-Nordic, line 293: You seem to meant ignoring CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND error but you still return it here?

src/crypto/PersistentStorageOperationalKeystore.cpp
```diff
@@ -86,6 +86,57 @@ CHIP_ERROR StoreOperationalKey(FabricIndex fabricIndex, PersistentStorageDelegat
...
+    VerifyOrReturnError(storage != nullptr, CHIP_ERROR_INVALID_ARGUMENT);
+    VerifyOrReturnError(IsValidFabricIndex(fabricIndex), CHIP_ERROR_INVALID_FABRIC_INDEX);
+
+    // Use a SensitiveDataBuffer to get RAII secret data clearing on scope exit.
+    Crypto::SensitiveDataBuffer<OpKeyTLVMaxSize()> buf;
+
+    // Load up the operational key structure from storage
+    uint16_t size  = static_cast<uint16_t>(buf.Capacity());
+    CHIP_ERROR err = storage->SyncGetKeyValue(DefaultStorageKeyAllocator::FabricOpKey(fabricIndex).KeyName(), buf.Bytes(), size);
+    if (err == CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND)
```
Damian-Nordic, line 101: In the PSA migration function you seem to assume that for already migrated fabric CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND will be returned, but due to this code this will likely be CHIP_ERROR_INVALID_FABRIC_INDEX, right?

src/crypto/PersistentStorageOperationalKeystore.cpp
```diff
@@ -310,4 +325,38 @@ void PersistentStorageOperationalKeystore::ReleaseEphemeralKeypair(Crypto::P256K
...
+{
+    VerifyOrReturnError(mStorage, CHIP_ERROR_INCORRECT_STATE);
+    VerifyOrReturnError(IsValidFabricIndex(fabricIndex), CHIP_ERROR_INVALID_FABRIC_INDEX);
+    VerifyOrReturnError(fabricIndex != kUndefinedFabricIndex, CHIP_ERROR_INVALID_FABRIC_INDEX);
+
+    P256SerializedKeypair serializedKeypair;
+
+    CHIP_ERROR err = operationalKeystore.ExportOpKeypairForFabric(fabricIndex, serializedKeypair);
+
+    if (CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND != err)
```
Damian-Nordic, line 339: Same comments as above: shouldn't you check _invalid fabric_ error instead, and should you return it instead of `CHIP_NO_ERROR`?

src/crypto/tests/TestPSAOpKeyStore.cpp
```diff
@@ -172,11 +179,82 @@ void TestEphemeralKeys(nlTestSuite * inSuite, void * inContext)
     opKeyStore.ReleaseEphemeralKeypair(ephemeralKeypair);
 }
 
+void TestMigrationKeys(nlTestSuite * inSuite, void * inContext)
+{
+    chip::TestPersistentStorageDelegate storage;
+    PSAOperationalKeystore psaOpKeyStore;
+    PersistentStorageOperationalKeystore persistentOpKeyStore;
+    FabricIndex kFabricIndex = 111;
```
Damian-Nordic, line 187:
```suggestion
    constexpr FabricIndex kFabricIndex = 111;
```

src/crypto/tests/TestPSAOpKeyStore.cpp
```diff
@@ -172,11 +179,82 @@ void TestEphemeralKeys(nlTestSuite * inSuite, void * inContext)
...
+    // Check what happens when we want to move a key of the specific Fabric, but the key does not exist.
+    NL_TEST_ASSERT(inSuite, storage.GetNumKeys() == 0);
+    NL_TEST_ASSERT(inSuite,
+                   psaOpKeyStore.MigrateOpKeypairForFabric(kFabricIndex, persistentOpKeyStore) == CHIP_ERROR_INVALID_FABRIC_INDEX);
+
+    auto generateAndStore = [&](FabricIndex index, MutableByteSpan & buf, P256PublicKey & pubKey) {
+        NL_TEST_ASSERT(inSuite, persistentOpKeyStore.HasPendingOpKeypair() == false);
+        NL_TEST_ASSERT(inSuite, persistentOpKeyStore.NewOpKeypairForFabric(kFabricIndex, buf) == CHIP_NO_ERROR);
+        NL_TEST_ASSERT(inSuite, VerifyCertificateSigningRequest(buf.data(), buf.size(), pubKey) == CHIP_NO_ERROR);
+        NL_TEST_ASSERT(inSuite, pubKey.Matches(pubKey));
```
Damian-Nordic, line 213: This doesn't check anything useful.

src/crypto/tests/TestPersistentStorageOpKeyStore.cpp
```diff
@@ -33,6 +33,99 @@ using namespace chip::Crypto;
...
+
+        return CHIP_NO_ERROR;
+    }
+
+    CHIP_ERROR ExportOpKeypairForFabric(FabricIndex fabricIndex, Crypto::P256SerializedKeypair & outKeypair) override
+    {
+        // Simulate not existing value
+        if (fabricIndex != mUsedFabricIndex)
+        {
+            return CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND;
```
Damian-Nordic, line 79: According to your doc in `OperationalKeystore.h`, this should return CHIP_ERROR_INVALID_FABRIC_INDEX.

Damian-Nordic (commented):
Good work with tests. Found some inconsistencies that you can look into.

src/crypto/PersistentStorageOperationalKeystore.cpp
ArekBalysNordic: I've changed it to return the CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND error for the Export method and keep CHIP_ERROR_INVALID_FABRIC_INDEX for the Signing method

Damian-Nordic (approved):

PR #376, opened 2024-01-10
[nrf noup] [zephyr] Fix CHIPDevicePlatformEvent.h include dependencies

markaj-nordic:
Currently the CHIPDevicePlatformEvent depends on the SystemPacketBuffer which is included in the
CHIPDeviceEvent.h too late.

The problem is silently worked around in the application when the <platform/PlatformManager.h> include is preceded by other headers that pull in SystemPacketBuffer.h, but we need a proper fix.

### DIFF
--- a/src/platform/Zephyr/CHIPDevicePlatformEvent.h
+++ b/src/platform/Zephyr/CHIPDevicePlatformEvent.h
@@ -23,7 +23,7 @@
 
 #pragma once
 
-#include <platform/CHIPDeviceEvent.h>
+#include <system/SystemPacketBuffer.h>
 
 #include <zephyr/bluetooth/bluetooth.h>
 


kkasperczyk-no (approved):

src/platform/Zephyr/CHIPDevicePlatformEvent.h
```diff
@@ -23,7 +23,7 @@
 
 #pragma once
 
-#include <platform/CHIPDeviceEvent.h>
+#include <system/SystemPacketBuffer.h>
```
ArekBalysNordic, line 26: Shouldn't we add it to the Matter upstream also?

kkasperczyk-no: I think we should, but probably the change is so small and non-controversial that it could be pushed upstream, as a part of 2.6.0 bump.

markaj-nordic: Yes, this was my initial thought as well.

ArekBalysNordic (approved):

PR #377, opened 2024-01-11
Fix the way of providing data to generate factory data

doublemis1:
Update the generation of factory data for nrfconnect platform. No more possible to provide spake2 verifier, always generated by provided passcode.

### DIFF
--- a/scripts/tools/nrfconnect/generate_nrfconnect_chip_factory_data.py
+++ b/scripts/tools/nrfconnect/generate_nrfconnect_chip_factory_data.py
@@ -99,8 +99,6 @@ def gen_test_certs(chip_cert_exe: str,
                    vendor_id: int,
                    product_id: int,
                    device_name: str,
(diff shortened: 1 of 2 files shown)


scripts/tools/nrfconnect/generate_nrfconnect_chip_factory_data.py
```diff
@@ -99,8 +99,6 @@ def gen_test_certs(chip_cert_exe: str,
                    vendor_id: int,
                    product_id: int,
                    device_name: str,
-                   generate_cd: bool = False,
```
ArekBalysNordic, line 102: I'm not sure that we can remove that, because some customers can use it to generate their CD. That option doesn't include CD in the FD set, but it generates CD to the build directory.

ArekBalysNordic:
Closing, all tasks from here have been done in: https://github.com/nrfconnect/sdk-connectedhomeip/pull/392

PR #378, opened 2024-01-12
[nrf noup] Added DNS server refresh after adding new IPv6 prefix

kkasperczyk-no:
The Wi-Fi device does not update mDNS queries after obtaining new IPv6 GUA prefix, so for some time after assigning prefix, the Thread Border Routers still use cached link-local address, which is not routable.

### DIFF
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -136,12 +136,13 @@ const Map<wifi_iface_state, WiFiManager::StationStatus, 10>
                               { WIFI_STATE_GROUP_HANDSHAKE, WiFiManager::StationStatus::PROVISIONING },
                               { WIFI_STATE_COMPLETED, WiFiManager::StationStatus::FULLY_PROVISIONED } });
 
-const Map<uint32_t, WiFiManager::NetEventHandler, 5>
-    WiFiManager::sEventHandlerMap({ { NET_EVENT_WIFI_SCAN_RESULT, WiFiManager::ScanResultHandler },
-                                    { NET_EVENT_WIFI_SCAN_DONE, WiFiManager::ScanDoneHandler },
-                                    { NET_EVENT_WIFI_CONNECT_RESULT, WiFiManager::ConnectHandler },
-                                    { NET_EVENT_WIFI_DISCONNECT_RESULT, WiFiManager::DisconnectHandler },
-                                    { NET_EVENT_WIFI_DISCONNECT_COMPLETE, WiFiManager::DisconnectHandler } });
+const Map<uint32_t, WiFiManager::NetEventHandler, 5> WiFiManager::sEventHandlerMap({
+    { NET_EVENT_WIFI_SCAN_RESULT, WiFiManager::ScanResultHandler },
+    { NET_EVENT_WIFI_SCAN_DONE, WiFiManager::ScanDoneHandler },
+    { NET_EVENT_WIFI_CONNECT_RESULT, WiFiManager::ConnectHandler },
+    { NET_EVENT_WIFI_DISCONNECT_RESULT, WiFiManager::DisconnectHandler },
+    { NET_EVENT_WIFI_DISCONNECT_COMPLETE, WiFiManager::DisconnectHandler },
+});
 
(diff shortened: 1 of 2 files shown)


src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -469,6 +480,19 @@ void WiFiManager::DisconnectHandler(Platform::UniquePtr<uint8_t>)
     });
 }
 
+void WiFiManager::IPv6PrefixAddHandler(const void * data)
+{
+    // This is needed to send mDNS queries containing new IPv6 prefix
+    chip::DeviceLayer::ChipDeviceEvent event;
```
markaj-nordic, line 486: nit: probably we can do 
```suggestion
   ChipDeviceEvent event {.Type = DeviceEventType::kDnssdRestartNeeded};
```
(similarly for kDnssdInitialized to be consistent)
And I would consider using namespace ::chip::DeviceLayer to have less namespaces in place.

markaj-nordic (approved):

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -469,6 +480,19 @@ void WiFiManager::DisconnectHandler(Platform::UniquePtr<uint8_t>)
...
+    event.Type = chip::DeviceLayer::DeviceEventType::kDnssdRestartNeeded;
+
+    CHIP_ERROR error = chip::DeviceLayer::PlatformMgr().PostEvent(&event);
+    if (error != CHIP_NO_ERROR)
+    {
+        ChipLogError(DeviceLayer, "Cannot post event [error: %s]", ErrorStr(error));
```
Damian-Nordic, line 492:
```suggestion
        ChipLogError(DeviceLayer, "Cannot post event: %" CHIP_ERROR_FORMAT, error.Format());
```
We should avoid using `ErrorStr` directly - `CHIP_ERROR_FORMAT` has lower overhead when error strings are disabled.

src/platform/nrfconnect/wifi/WiFiManager.cpp
Damian-Nordic: The code is already inside `DeviceLayer` so not even "using namespace" is needed

src/platform/nrfconnect/wifi/WiFiManager.h
```diff
@@ -198,14 +198,18 @@ class WiFiManager
     constexpr static uint32_t kWifiManagementEvents = NET_EVENT_WIFI_SCAN_RESULT | NET_EVENT_WIFI_SCAN_DONE |
         NET_EVENT_WIFI_CONNECT_RESULT | NET_EVENT_WIFI_DISCONNECT_RESULT | NET_EVENT_WIFI_IFACE_STATUS;
 
+    constexpr static uint32_t kIPv6ManagementEvents = NET_EVENT_IPV6_PREFIX_ADD;
```
Damian-Nordic, line 201: Wouldn't it be more straightforward to react on `NET_EVENT_IPV6_ADDR_ADD` and `NET_EVENT_IPV6_ADDR_DEL`? That way, this would be correctly handled even if the address be added via other means. Then you could filter out link-local addresses in the event handler.

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Ahh I mistakenly assumed it's only scoped to ToScanResponse.

src/platform/nrfconnect/wifi/WiFiManager.h
markaj-nordic: Makes sense to me. Assuming that for whatever reason these are not excluding addresses assigned by SLAAC (I believe this eventing should be "source" agnostic but still).

kkasperczyk-no: > Wouldn't it be more straightforward to react on NET_EVENT_IPV6_ADDR_ADD and NET_EVENT_IPV6_ADDR_DEL?
Ok, actually that was my first idea. Then I decided to use `NET_EVENT_IPV6_PREFIX_ADD` to skip the link-local address filtering part, but I get your point.

Damian-Nordic (approved):

PR #380, opened 2024-01-19
[nrf noup] Lock thread stack before factory reset

LipinskiPNordicSemi:
This commit fix a problem with thread activity interrupting factory reset. This activity used to led to “Factory reset fail: -6”. Writing to cleared nvm flash pages caused the problem.

### DIFF
--- a/src/platform/Zephyr/ConfigurationManagerImpl.cpp
+++ b/src/platform/Zephyr/ConfigurationManagerImpl.cpp
@@ -40,6 +40,10 @@
 #include <zephyr/settings/settings.h>
 #endif
 
+#ifdef CONFIG_NET_L2_OPENTHREAD
+#include <platform/ThreadStackManager.h>
+#endif 
+
 namespace chip {
 namespace DeviceLayer {
 
@@ -176,6 +180,11 @@ void ConfigurationManagerImpl::RunConfigUnitTest(void)
 void ConfigurationManagerImpl::DoFactoryReset(intptr_t arg)
 {
     ChipLogProgress(DeviceLayer, "Performing factory reset");
+
+// Lock the Thread stack to avoid unwanted interaction with settings NVS during factory reset.
+#ifdef CONFIG_NET_L2_OPENTHREAD
+    ThreadStackMgr().LockThreadStack();
+#endif
 
 #ifdef CONFIG_CHIP_FACTORY_RESET_ERASE_NVS
     void * storage = nullptr;


src/platform/Zephyr/ConfigurationManagerImpl.cpp
```diff
@@ -176,6 +176,7 @@ void ConfigurationManagerImpl::RunConfigUnitTest(void)
 void ConfigurationManagerImpl::DoFactoryReset(intptr_t arg)
 {
     ChipLogProgress(DeviceLayer, "Performing factory reset");
+    ThreadStackManager().LockThreadStack();
```
Damian-Nordic, line 179: This probably won't compile (?)
```suggestion
    ThreadStackMgr().LockThreadStack();
```

Also, do not assume it's Matter over **Thread**.

LipinskiPNordicSemi: @Damian-Nordic Is there any Matter api which allow to disable thread?

Damian-Nordic: Not sure I understand, I meant that you should compile this line only if `CONFIG_NET_L2_OPENTHREAD` is defined as it may not compile for Wi-Fi.

Damian-Nordic: Btw. you could add a comment explaining why this is needed as `Lock` without `Unlock` can look weird to newcomers :).

src/platform/Zephyr/ConfigurationManagerImpl.cpp
```diff
@@ -176,6 +176,7 @@ void ConfigurationManagerImpl::RunConfigUnitTest(void)
 void ConfigurationManagerImpl::DoFactoryReset(intptr_t arg)
 {
     ChipLogProgress(DeviceLayer, "Performing factory reset");
+    ThreadStackMgr().LockThreadStack();
```
markaj-nordic, line 179: This must be guarded with  `#ifdef CONFIG_NET_L2_OPENTHREAD`
```suggestion
#ifdef CONFIG_NET_L2_OPENTHREAD
    ThreadStackMgr().LockThreadStack();
#endif
```
Also, as Damian suggested, please describe what is the purpose of this call here and why it can be used here without unlocking the stack.

markaj-nordic (changes requested):

src/platform/Zephyr/ConfigurationManagerImpl.cpp
```diff
@@ -177,6 +177,11 @@ void ConfigurationManagerImpl::DoFactoryReset(intptr_t arg)
 {
     ChipLogProgress(DeviceLayer, "Performing factory reset");
 
+// Lock the Thread to not interupt in settings during factory reset
```
markaj-nordic, line 180:
```suggestion
// Lock the Thread stack to avoid unwanted interaction with settings NVS during factory reset.
```

markaj-nordic (approved):

Damian-Nordic (approved):

PR #381, opened 2024-01-26
[nrf noup] nrfconnect/telink: Adapt to zcbor 0.8.1

oyvindronningstad:
New parameter to ZCBOR_STATE_D()


kkasperczyk-no (approved):
I assume that new flag argument set to 0 has no impact on current functionality, so approving.

LuDuda (approved):

PR #383, opened 2024-02-09
Hkdf integration

kkasperczyk-no:
Pulled the most recent changes in CHIPCryptoPAL to integrate usage of HKDF key handle instead of raw data.


src/crypto/PSASpake2p.cpp
```diff
@@ -179,24 +179,23 @@ CHIP_ERROR PSASpake2p_P256_SHA256_HKDF_HMAC::KeyConfirm(const uint8_t * in, size
...
+     * The currently used PSA version does not provide API to get shared key yet, however it is present on the most recent version.
+     * After it will be integrated, the extracting raw data from oberon context could be replaced with the following code:
+     *
+     *  HkdfKeyAttributes attrs;
+     *  psa_status_t status = psa_pake_get_shared_key(&mOperation, attrs.Get(), &key);
      */
-    oberon_spake2p_operation_t & oberonCtx = mOperation.MBEDTLS_PRIVATE(ctx).oberon_pake_ctx.ctx.oberon_spake2p_ctx;
-
-    VerifyOrReturnError((oberonCtx.hash_len / 2) <= *out_len, CHIP_ERROR_BUFFER_TOO_SMALL);
+    const oberon_spake2p_operation_t & oberonCtx = mOperation.MBEDTLS_PRIVATE(ctx).oberon_pake_ctx.ctx.oberon_spake2p_ctx;
```
LuDuda, line 196: The overall idea for refactoring the code by Damian, was to being able to remove this line, and avoid using `psa_pake_get_shared_key` (which is not yet introduced). I checked with Damian, that this will be possible :)

LuDuda (changes requested):

src/crypto/PSASpake2p.cpp
ArekBalysNordic: Ok, so we need to import a shared key to PSA at the end of the spake algorithm and assign it's key reference in this method. Anyway, we can use the psa_pake_get_shared_key once it is available so we can leave a comment there.

LuDuda:
As discussed F2F, this PR could come without the requested optimization. So for a time being, removing request change.

src/crypto/PSASessionKeystore.cpp
```diff
@@ -185,10 +185,17 @@ void PSASessionKeystore::DestroyKey(Symmetric128BitsKeyHandle & key)
...
-    psa_destroy_key(keyId);
-    keyId = 0;
+    if (keyHandle.mIsKeyId)
+    {
+        psa_destroy_key(keyHandle.mKeyId);
+        keyHandle.mKeyId = 0;
+    }
+    else
+    {
+        Platform::Delete(keyHandle.mKeyDerivationOp);
```
Damian-Nordic, line 197: mKeyDerivationOp = nullptr to satisfy the API contract?

src/crypto/PSASpake2p.cpp
```diff
@@ -179,22 +180,27 @@ CHIP_ERROR PSASpake2p_P256_SHA256_HKDF_HMAC::KeyConfirm(const uint8_t * in, size
...
     /*
-     * TODO: either:
-     * - use psa_pake_shared_secret() proposed in https://github.com/ARM-software/psa-api/issues/86
-     * - refactor Matter's GetKeys API to take an abstract shared secret instead of raw secret bytes.
+     * TODO: use psa_pake_shared_secret() proposed in https://github.com/ARM-software/psa-api/issues/86
      */
-    oberon_spake2p_operation_t & oberonCtx = mOperation.MBEDTLS_PRIVATE(ctx).oberon_pake_ctx.ctx.oberon_spake2p_ctx;
 
-    VerifyOrReturnError((oberonCtx.hash_len / 2) <= *out_len, CHIP_ERROR_BUFFER_TOO_SMALL);
+    psa_key_derivation_operation_t * kdf = Platform::New<psa_key_derivation_operation_t>();
```
Damian-Nordic, line 189: Use UniquePtr to avoid memory leaks if one of the PSA operations below fails.

Damian-Nordic (approved):
Reviewed the last commit only. Approved provided the 3 issues are fixed.

src/platform/nrfconnect/CHIPPlatformConfig.h
```diff
@@ -40,7 +40,8 @@
 
 #ifdef CONFIG_CHIP_CRYPTO_PSA
 #define CHIP_CONFIG_SHA256_CONTEXT_SIZE sizeof(psa_hash_operation_t)
-#define CHIP_CONFIG_HKDF_KEY_HANDLE_CONTEXT_SIZE sizeof(psa_key_id_t)
+// Alignment to sizeof(PsaHkdfKeyHandle) from crypto/CHIPCryptoPALPSA.h.
+#define CHIP_CONFIG_HKDF_KEY_HANDLE_CONTEXT_SIZE (sizeof(psa_key_id_t) + sizeof(bool))
```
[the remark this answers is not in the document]
kkasperczyk-no: Right, but is it a problem? Or you mean updating the comment to be more precise?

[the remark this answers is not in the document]
Damian-Nordic: A bit confusing because `CHIP_CONFIG_HKDF_KEY_HANDLE_CONTEXT_SIZE` is 5 and we cast the handle to a structure that is 8 bytes. But maybe I'm too paranoid so let's leave it as it is :).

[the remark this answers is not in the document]
kkasperczyk-no: Ah, now I see what you mean. Yeah, indeed it looks incorrect. The best option would be to use the `sizeof(PsaHkdfKeyHandle)`, but include is not possible. Another option could be to declare the structure of the same content, as `PsaHkdfKeyHandle`, but I'm not sure. Maybe for now it's ok to leave it, as we are going to revert it soon.

src/crypto/CHIPCryptoPALPSA.h
```diff
@@ -28,6 +28,7 @@
 
 #include "CHIPCryptoPAL.h"
 #include <lib/core/DataModelTypes.h>
+#include <lib/support/CHIPMem.h>
```
ArekBalysNordic, line 31: Do we need it here?

src/crypto/PSASpake2p.h
```diff
@@ -144,14 +145,14 @@ class PSASpake2p_P256_SHA256_HKDF_HMAC
     CHIP_ERROR KeyConfirm(const uint8_t * in, size_t in_len);
 
     /**
-     * @brief Return the shared secret.
+     * @brief Return the shared HKDF key.
```
ArekBalysNordic, line 148: Maybe `Return the HKDF Key handle containing a key reference to the shared secret`?

ArekBalysNordic (approved):
Two minor comments, overall LGTM : :)

src/crypto/CHIPCryptoPALPSA.h
kkasperczyk-no: No, good catch, it's a leftover after something that I tried to do here before :)

src/crypto/PSASpake2p.h
kkasperczyk-no: Ok, I just copied the docs from `CHIPCryptoPALPSA::GetKeys`, but extending it can be valuable.

src/crypto/CHIPCryptoPALPSA.h
```diff
@@ -145,11 +156,11 @@ class PsaKdf
     CHIP_ERROR DeriveKey(const psa_key_attributes_t & attributes, psa_key_id_t & keyId);
 
 private:
-    CHIP_ERROR InitOperation(psa_key_id_t hkdfKey, const ByteSpan & salt, const ByteSpan & info);
+    CHIP_ERROR InitOperation(PsaHkdfKeyHandle hkdfKey, const ByteSpan & salt, const ByteSpan & info);
 
-    psa_key_id_t mSecretKeyId                 = 0;
-    psa_key_derivation_operation_t mOperation = PSA_KEY_DERIVATION_OPERATION_INIT;
+    psa_key_id_t mSecretKeyId                             = 0;
```
LuDuda, line 161: Can we use `PSA_KEY_ID_NULL` instead of 0, or there are some issues with it?

src/crypto/PSASpake2p.cpp
```diff
@@ -179,22 +180,30 @@ CHIP_ERROR PSASpake2p_P256_SHA256_HKDF_HMAC::KeyConfirm(const uint8_t * in, size
...
+    status = psa_pake_get_implicit_key(&mOperation, kdfPtr.get());
+    VerifyOrReturnError(status == PSA_SUCCESS, CHIP_ERROR_INTERNAL);
+
+    auto & hkdfKeyHandle           = key.AsMutable<PsaHkdfKeyHandle>();
+    hkdfKeyHandle.mKeyDerivationOp = kdfPtr.get();
+    hkdfKeyHandle.mIsKeyId         = false;
 
-    memcpy(out, oberonCtx.shared, oberonCtx.hash_len / 2);
-    *out_len = oberonCtx.hash_len / 2;
+    kdfPtr.release();
```
LuDuda, line 206: I guess it's fine and more clean.. but do we need to release it explicitly?

src/platform/nrfconnect/CHIPPlatformConfig.h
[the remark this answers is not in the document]
LuDuda: If it works, then fine! But don't we have any macro or util to align to 8-bytes in Matter?

LuDuda (approved):
Looks solid! Good work.

src/crypto/PSASpake2p.cpp
kkasperczyk-no: Could you clarify? We have to release it to inform the UniquePtr that the memory will be deleted in some other place (by the `DestroyKey` method). If we won't do that, the UniquePtr will delete key just after we leave the scope.

src/crypto/CHIPCryptoPALPSA.h
kkasperczyk-no: Sure, we can do that.

src/platform/nrfconnect/CHIPPlatformConfig.h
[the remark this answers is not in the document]
kkasperczyk-no: There are cpp macros to generally align the data structure, or return the aligned size of specific type, but to be honest I don't know the macro that would return aligned size from the numerical unaligned size.

PR #389, opened 2024-02-16
[nrf noup] [zephyr] Allow BLE advertising restarts

markaj-nordic:
* allow BLE advertising restarts in case of failures
* which can be triggered by calling SetBLEAdvertisingEnabled(true) ConnectivityMgr public API from the application code
* do not register CHIPoBLE GATT services when the advertising cannot be started
* this allows the disconnection handler to filter out non-Matter BLE connections that were terminated
* fix possible underflow of connection counters

### DIFF
--- a/src/platform/Zephyr/BLEManagerImpl.cpp
+++ b/src/platform/Zephyr/BLEManagerImpl.cpp
@@ -161,7 +161,7 @@ BLEManagerImpl BLEManagerImpl::sInstance;
 CHIP_ERROR BLEManagerImpl::_Init()
 {
     int err = 0;
-    int id = 0;
+    int id  = 0;
 
     mServiceMode = ConnectivityManager::kCHIPoBLEServiceMode_Enabled;
     mFlags.ClearAll().Set(Flags::kAdvertisingEnabled, CHIP_DEVICE_CONFIG_CHIPOBLE_ENABLE_ADVERTISING_AUTOSTART);
@@ -240,7 +240,13 @@ void BLEManagerImpl::DriveBLEState()
         {
             mFlags.Clear(Flags::kAdvertisingRefreshNeeded);
             err = StartAdvertising();
-            SuccessOrExit(err);
+            if (err != CHIP_NO_ERROR)
+            {
+                // Return prematurely but keep the CHIPoBLE service mode enabled to allow advertising retries
+                mServiceMode = ConnectivityManager::kCHIPoBLEServiceMode_Enabled;
+                ChipLogError(DeviceLayer, "Could not start CHIPoBLE service due to error: %" CHIP_ERROR_FORMAT, err.Format());
+                return;
+            }
         }
     }
     else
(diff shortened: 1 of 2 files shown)


src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -240,15 +240,26 @@ void BLEManagerImpl::DriveBLEState()
...
     else
     {
         if (mFlags.Has(Flags::kAdvertising))
         {
             err = StopAdvertising();
-            SuccessOrExit(err);
+            if (err != CHIP_NO_ERROR)
+            {
+                ChipLogError(DeviceLayer, "Disabling CHIPoBLE service due to error: %" CHIP_ERROR_FORMAT, err.Format());
+                mServiceMode = ConnectivityManager::kCHIPoBLEServiceMode_Disabled;
```
kkasperczyk-no, line 260: Shouldn't we call bt_gatt_service_unregister after that? Generally maybe we could use UnregisterGattService below to not duplicate the code.

src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -240,15 +240,26 @@ void BLEManagerImpl::DriveBLEState()
         {
             mFlags.Clear(Flags::kAdvertisingRefreshNeeded);
             err = StartAdvertising();
-            SuccessOrExit(err);
+            if (err != CHIP_NO_ERROR)
+            {
+                // Return prematurely but keep the CHIPoBLE service mode enabled to allow advertising retries
```
kkasperczyk-no, line 245: I think I don't understand. Is there any other place in the code that calls `StartAdvertising` (I can't find it)? If no, it looks like we unregister Gatt service in the `StartAdvertising` method in case of failure, but here we enable the service back. Is there something I'm missing?

markaj-nordic: If we disable the service mode here (as it used to be done before my change) we will not be able to drive the BLE state to reach the `StartAdvertising()` in line 242 after the previous advertising request failed. So, this modification allows to do the actual restart. In `StartAdvertising()` we unregister the service but we do not disable the service mode (please take a look at the description of `kChipoBleGattServiceRegister flag`). Semantically we apply the `kAdvertisingEnabled/kAdvertising` scheme to the BLE service being enabled/registered (`kCHIPoBLEServiceMode_*abled/kChipoBleGattServiceRegister`). On the other hand maybe we may consider to stop using service mode concept at all, as it seems redundant to me, and retry to advertise regardless of it (but maybe I'm not aware of some edge cases).

src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -240,15 +240,26 @@ void BLEManagerImpl::DriveBLEState()
...
+                mServiceMode = ConnectivityManager::kCHIPoBLEServiceMode_Enabled;
```
ArekBalysNordic, line 246: This is the case when an error occurs, right? So why do we set mServiceMode to Enabled?

src/platform/Zephyr/BLEManagerImpl.cpp
kkasperczyk-no: Discussed F2F, my comments are not relevant.

kkasperczyk-no (approved):

src/platform/Zephyr/BLEManagerImpl.cpp
markaj-nordic: This is to keep the service mode globally enabled to be able to restart the advertising after it fails. This is how the DriveBLEState logic works. But we do clear the kChipoBleGattServiceRegister and unregister the service in case of a failure.

ArekBalysNordic (approved):

PR #390, opened 2024-02-19
Fixed PlatformMgr().Shutdown() for NRF53

bdlr2:
Linked to my case on the DevZone https://devzone.nordicsemi.com/f/nordic-q-a/108101/device-crash-when-doing-a-software-reset

When calling the function `PlatformMgr().Shutdown();` (intentionally, or through a factory reset) a Hard Fault happens because the communication is broken/not in sync anymore between the App and Net cores of the NRF5340. My fix proposes to first reset the network core then reset the application core.

### DIFF
--- a/src/include/platform/internal/GenericPlatformManagerImpl_Zephyr.ipp
+++ b/src/include/platform/internal/GenericPlatformManagerImpl_Zephyr.ipp
@@ -42,6 +42,10 @@
 #include <psa/crypto.h>
 #endif
 
+#if CONFIG_SOC_NRF5340_CPUAPP
+    #include <hal/nrf_reset.h>
+#endif
+
 #define DEFAULT_MIN_SLEEP_PERIOD (60 * 60 * 24 * 30) // Month [sec]
 
 namespace chip {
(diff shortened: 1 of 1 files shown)


LuDuda (changes requested):
Hello @bdlr2,

Thank you for your contribution.

I believe we should not need to call `nrf_reset_network_force_off` in the Matter platform, as this should be handled internally by `sys_reboot`.

Have you encountered this issue on the nRF5340DK, or only on a custom board? If I had to guess, it's possible you've designed your own board and the `nrf5340_cpunet_reset.c` file has not been included in board's `CMakeLists.txt`. You can find an example here (note the if-statement): [nRF5340DK CMakeLists.txt](https://github.com/nrfconnect/sdk-zephyr/blob/v3.5.99-ncs1/boards/arm/nrf5340dk_nrf5340/CMakeLists.txt#L4-L7).

bdlr2:
Hello @LuDuda, thanks for the feedback.
That's indeed on a custom board, but I double checked and the `nrf5340_cpunet_reset.c` is included properly in the `CMakelists.txt`. My board was derived from the nRF5340DK to avoid this kind of issues.

LuDuda:
> Hello @LuDuda, thanks for the feedback. That's indeed on a custom board, but I double checked and the `nrf5340_cpunet_reset.c` is included properly in the `CMakelists.txt`. My board was derived from the nRF5340DK to avoid this kind of issues.

@bdlr2 Have you been able to check the newest SDK if the problem still exists? As i mentioned above, I would like to fix this issue outside of the Matter code.

bdlr2:
@LuDuda no, I didn't check yet.

PR #393, opened 2024-02-23
[nrf fromlist][nrfconnect] Add checking of Wi-Fi supplicant and Interface

ArekBalysNordic:
We need to handle the Wi-Fi supplicant events to verify whether the Wi-Fi supplicant and Interface are ready to use.

### DIFF
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -144,6 +144,33 @@ const Map<uint32_t, WiFiManager::NetEventHandler, 5> WiFiManager::sEventHandlerM
     { NET_EVENT_WIFI_DISCONNECT_COMPLETE, WiFiManager::DisconnectHandler },
 });
 
+void WiFiManager::WiFiSupplicantEventHandler(net_mgmt_event_callback * cb, uint32_t mgmtEvent, net_if * iface)
+{
+    switch (mgmtEvent)
+    {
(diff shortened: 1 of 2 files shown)


src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -144,6 +144,33 @@ const Map<uint32_t, WiFiManager::NetEventHandler, 5> WiFiManager::sEventHandlerM
     { NET_EVENT_WIFI_DISCONNECT_COMPLETE, WiFiManager::DisconnectHandler },
 });
 
+void WiFiManager::WiFiSupplicantEventHandler(net_mgmt_event_callback * cb, uint32_t mgmtEvent, net_if * iface)
+{
+    switch (mgmtEvent)
+    {
+    case NET_EVENT_WPA_SUPP_READY:
```
markaj-nordic, line 151: Is this possible that any of these events is generated after we successfully initialized supplicant/iface? I wonder if it would make any sense to process them and inform an application that something went wrong.

markaj-nordic (approved):

src/platform/nrfconnect/wifi/WiFiManager.cpp
ArekBalysNordic: I thought about that, but it seems that we need to wait ~50-100 ms to receive those events after initializing our WiFi Manager.

LuDuda (approved):

PR #394, opened 2024-02-23
[nrf noup] [nrfconnect] [zephyr] Disable synchronous printk

markaj-nordic:
Disable synchronous printk to avoid blocking IRQs which may affect time sensitive components (like 15.4 radio).

### DIFF
--- a/config/zephyr/chip-module/Kconfig.defaults
+++ b/config/zephyr/chip-module/Kconfig.defaults
@@ -40,9 +40,11 @@ config LOG_DEFAULT_LEVEL
 
 endif
 
+# disable synchronous printk to avoid blocking IRQs which
+# may affect time sensitive components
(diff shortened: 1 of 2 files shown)


Damian-Nordic (approved):

config/zephyr/chip-module/Kconfig.defaults
```diff
@@ -40,9 +40,11 @@ config LOG_DEFAULT_LEVEL
 
 endif
 
+# disable synchronous printk to avoid blocking IRQs which
```
Damian-Nordic, line 43: fyi: we don't use this file (it's for standalone zephyr platform in Matter)

markaj-nordic: Yes, I know. I thought i would be more consistent anyway.

PR #398, opened 2024-02-29
[nrf noup] Increase CHIP stack size

LuDuda:
This commit increases the current CHIP stack size to allow building with LTO and provide more buffer for non-LTO compilation.


src/platform/nrfconnect/CHIPDevicePlatformConfig.h
```diff
@@ -215,7 +215,7 @@
 #endif // CHIP_DEVICE_CONFIG_CHIP_TASK_PRIORITY
 
 #ifndef CHIP_DEVICE_CONFIG_CHIP_TASK_STACK_SIZE
-#define CHIP_DEVICE_CONFIG_CHIP_TASK_STACK_SIZE 6144
+#define CHIP_DEVICE_CONFIG_CHIP_TASK_STACK_SIZE 8192
```
Damian-Nordic, line 218: nit: it could be `if CONFIG_LTO then 8 else 6`, right?

Damian-Nordic (approved):

markaj-nordic (approved):

Damian-Nordic (approved):

PR #399, opened 2024-03-01
[nrf fromtree] Clear srp implementation

LipinskiPNordicSemi:
This commit implements the host srp clear, using a cherry-picked upstream solution.

Cherry-picked commit 6350333 content:

“This PR brings https://github.com/project-chip/connectedhomeip/pull/31606 feature to The GenericThreadStackManagerImpl.
The functionality remains the same (with the addition of deleting the client key lease also).

However, this requires each platform that uses it, to provide a synchronization mechanism.

I created the Templates _WaitOnSrpClearAllComplete and _NotifySrpClearAllComplete and I provided an implementation for all platforms using the GenericThreadStackManagerImpl. (It is left unimplemented for others)

The synchronization mechanism I chose for Zephyr platforms (nordic/telink) and Infineon might not be optional so suggestions are welcome.

The functionality was only tested on the Silabs platform and is currently only used on it also. I am confident it works for all FreeRTOS-based platforms.

I suggest all other thread platform owners take some time to test and use ClearAllSrpHostAndServices during their factory-reset sequence.”

### DIFF
--- a/src/platform/Zephyr/ConfigurationManagerImpl.cpp
+++ b/src/platform/Zephyr/ConfigurationManagerImpl.cpp
@@ -42,7 +42,7 @@
 
 #ifdef CONFIG_NET_L2_OPENTHREAD
 #include <platform/ThreadStackManager.h>
-#endif 
+#endif
 
 namespace chip {
 namespace DeviceLayer {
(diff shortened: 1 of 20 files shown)


src/platform/Zephyr/ConfigurationManagerImpl.cpp
```diff
@@ -181,6 +181,12 @@ void ConfigurationManagerImpl::DoFactoryReset(intptr_t arg)
 {
     ChipLogProgress(DeviceLayer, "Performing factory reset");
 
+#if CHIP_DEVICE_CONFIG_ENABLE_THREAD_SRP_CLIENT
+    ThreadStackMgr().WaitOnSrpClearAllComplete();
```
Damian-Nordic, line 185: Why isn't `ThreadStackMgr().ClearAllSrpHostAndServices()` enough? The other functions are called under the hood.

LipinskiPNordicSemi: I didn't check that, my bad

Damian-Nordic (approved):

LuDuda:
Thanks. What about manifest update?

markaj-nordic:
This patch did not land in NCS 2.6.0. I will be pulled in by the next connectedhomeip upmerge. Closing.

PR #400, opened 2024-03-05
[nrf noup] Fix the build warning in door-lock-server

LipinskiPNordicSemi:
This fixes the uninitialized variable warning appearing when compiling the door lock app.


LuDuda (approved):

markaj-nordic (approved):
I'm still curious why this fix landed only in the 2.5.0 release.

PR #401, opened 2024-03-06
[nrf fromlist] [app] Do not crash if CommandHandler fails to allocate packet

Damian-Nordic:
CommandHandler uses VerifyOrDie when adding a status to be sent to the requestor. If the device runs out packet buffers and CommandHandler fails to allocate a packet for the status, the device crashes.

Triggering the crash requires many commands to arrive in the device around the same time, which is probably rare but possible.

### DIFF
--- a/src/app/CommandHandler.cpp
+++ b/src/app/CommandHandler.cpp
@@ -467,7 +467,17 @@ void CommandHandler::AddStatus(const ConcreteCommandPath & aCommandPath, const P
 {
     // Return prematurely in case of requests targeted to a group that should not add the status for response purposes.
     VerifyOrReturn(!IsGroupRequest());
-    VerifyOrDie(FallibleAddStatus(aCommandPath, aStatus, context) == CHIP_NO_ERROR);
+
+    CHIP_ERROR error = FallibleAddStatus(aCommandPath, aStatus, context);
+
+    if (error != CHIP_NO_ERROR)
+    {
+        ChipLogError(DataManagement, "Failed to add command status: %" CHIP_ERROR_FORMAT, error.Format());
+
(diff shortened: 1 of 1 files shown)


src/app/CommandHandler.cpp
```diff
@@ -467,7 +467,17 @@ void CommandHandler::AddStatus(const ConcreteCommandPath & aCommandPath, const P
...
+
+    CHIP_ERROR error = FallibleAddStatus(aCommandPath, aStatus, context);
+
+    if (error != CHIP_NO_ERROR)
+    {
+        ChipLogError(DataManagement, "Failed to add command status: %" CHIP_ERROR_FORMAT, error.Format());
+
+        // Do not crash if the status has not been added due to running out of packet buffers or other resources.
+        // It is better to drop a single response than to go offline and lose all sessions and subscriptions.
+        VerifyOrDie(error == CHIP_ERROR_NO_MEMORY);
```
markaj-nordic, line 479: So we crash only if the error is other than CHIP_ERROR_NO_MEMORY? Judging by the comment I expected to cash exactly in such a case. Or am I lost here? :)

Damian-Nordic: I think the comment is correct: Do **not** crash in the case of failure due to "running out of...". But if you have a proposal how to rephrase that to make it clearer for you, spit it out :).

LuDuda (approved):

src/app/CommandHandler.cpp
markaj-nordic: discussed f2f.

markaj-nordic (approved):

PR #403, opened 2024-03-07
[nrf noup] Select Experimental for persistent subscriptions

ArekBalysNordic:
Currently, there is an issue with persistent subscriptions when multiple controllers create subscriptions.

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig
+++ b/config/nrfconnect/chip-module/Kconfig
@@ -297,6 +297,10 @@ config CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
 	  and remove it. After the first boot of the device the DAC private key will be moved
 	  to the PSA ITS secure storage and will not be available in the factory data anymore.
 	  It will be overwritten in the factory data set by zeros.
-	
+
+config CHIP_PERSISTENT_SUBSCRIPTIONS
+	default n
+	# selecting experimental for this feature since there is an issue with multiple controllers.
+	select EXPERIMENTAL
 
 endif # CHIP


config/nrfconnect/chip-module/Kconfig.features
```diff
@@ -290,4 +290,9 @@ config CHIP_LAST_FABRIC_REMOVED_ACTION_DELAY
 	  an action chosen by the CHIP_LAST_FABRIC_REMOVED_ACTION option. This schedule will allow for
 	  avoiding race conditions before the device removes non-volatile data.
 
+config CHIP_PERSISTENT_SUBSCRIPTIONS
```
Damian-Nordic, line 293: Why is this in `Kconfig.features` and not `Kconfig`?

LuDuda (approved):

config/nrfconnect/chip-module/Kconfig.features
LuDuda: I thought about it as well when reviewing but the config/Zephyr where this is placed is generic for more platforms now. But I'm fine with adding experimental flag there - anyway.

ArekBalysNordic: It is the Matter feature and it is optional AFAIR, but I don't have a strong opinion on where it should be added. :)

ArekBalysNordic: Do you think that it fits better to Kconfig?

ArekBalysNordic: Ok, I see, indeed we may do something with that to make it more intuitive :)

Damian-Nordic (approved):

ArekBalysNordic:
Unit Test error is not related to this PR.

config/nrfconnect/chip-module/Kconfig.features
LuDuda: Cool, i misdunderstood, that we want to move it to `Zephyr` folder. Thanks, all makes sense.

PR #404, opened 2024-03-07
[nrf noup] fix various issues with Network Commissioning cluster on Wi-Fi

LuDuda:
This PR fixes few issues with Network Commissioning cluster on Wi-Fi platform.

Please review commit by commit.

**[nrf noup] Fix handling of LastNetworkID in Wi-Fi driver**
This commit ensures that `LastNetworkID` attribute is not null, not only after `ConnectNetwork` command is issued, but also in case the connection is established after resboot or after connection is lost.

`[1709852472.587357][369591:369593] CHIP:TOO:   LastNetworkID: 536368726F64696E676572205761766573203547`

**[nrf noup] Do not clear Last Network ID and Connection Error after scan**
This bug is already fixed upstream for Matter 1.3. This fixes the issue where `LastNetworkID` attribute is cleared after `ScanNetworks` command is completed.

**[nrf noup] Fix various Wi-Fi issues with error code handling**
This commit handles a few issues with Wi-Fi connection or scanning:
 - Use wifi_status structure instead of incompatible WiFiRequestStatus
 - On connect error value > 2 do not report success*
 - On scan error value > 1 do not report success
 - Provide value of mandatory LastConnectErrorValue attribute
 
*This is especially critical.

### DIFF
--- a/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
+++ b/src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
@@ -103,8 +103,7 @@ CHIP_ERROR NrfWiFiDriver::Init(NetworkStatusChangeCallback * networkStatusChange
 
     if (mStagingNetwork.IsConfigured())
     {
-        WiFiManager::ConnectionHandling handling{ [] { Instance().OnNetworkStatusChanged(Status::kSuccess); },
-                                                  [] { Instance().OnNetworkStatusChanged(Status::kUnknownError); },
+        WiFiManager::ConnectionHandling handling{ [](int connStatus) { Instance().OnNetworkStatusChanged(connStatus); },
                                                   System::Clock::Seconds32{ kWiFiConnectNetworkTimeoutSeconds } };
         ReturnErrorOnFailure(
             WiFiManager::Instance().Connect(mStagingNetwork.GetSsidSpan(), mStagingNetwork.GetPassSpan(), handling));
@@ -113,8 +112,10 @@ CHIP_ERROR NrfWiFiDriver::Init(NetworkStatusChangeCallback * networkStatusChange
     return CHIP_NO_ERROR;
 }
 
-void NrfWiFiDriver::OnNetworkStatusChanged(Status status)
+void NrfWiFiDriver::OnNetworkStatusChanged(int connStatus)
 {
+    Status status = connStatus ? Status::kUnknownError : Status::kSuccess;
+
     if (status == Status::kSuccess)
     {
         ConnectivityMgr().SetWiFiStationMode(ConnectivityManager::kWiFiStationMode_Enabled);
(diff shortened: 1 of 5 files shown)


src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -122,7 +122,13 @@ void NrfWiFiDriver::OnNetworkStatusChanged(Status status)
 
     if (mpNetworkStatusChangeCallback)
     {
-        mpNetworkStatusChangeCallback->OnNetworkingStatusChange(status, NullOptional, NullOptional);
+        WiFiManager::WiFiInfo wifiInfo;
+
+        if (CHIP_NO_ERROR == WiFiManager::Instance().GetWiFiInfo(wifiInfo))
```
markaj-nordic, line 127: Should we fallback to NullOptional in case the GetWiFiInfo() fails for whatever reason?

markaj-nordic (approved):
LGTM! Thanks.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
LuDuda: This might be okay, however I'm not sure. I see that in all 15 platforms the correct NetworkID is used, or `OnNetworkingStatusChange` is skipped.

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
```diff
@@ -113,8 +112,10 @@ CHIP_ERROR NrfWiFiDriver::Init(NetworkStatusChangeCallback * networkStatusChange
     return CHIP_NO_ERROR;
 }
 
-void NrfWiFiDriver::OnNetworkStatusChanged(Status status)
+void NrfWiFiDriver::OnNetworkStatusChanged(int connStatus)
 {
+    Status status = connStatus ? Status::kUnknownError : Status::kSuccess;
```
Damian-Nordic, line 117: Add a TODO to check if we can return more accurate statuses?

Damian-Nordic (approved):
It's definitely better than it was before so approving but added a few suggestions for future improvements :)

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
markaj-nordic: But don't we support only one network anyway? If the last connection failed then we can't interact with the device anyway, right?

Damian-Nordic: Yeah, but the connection may also fail during commissioning.

src/platform/nrfconnect/wifi/WiFiManager.h
```diff
@@ -88,16 +88,9 @@ class Map
...
-        SUCCESS    = 0,
-        FAILURE    = 1,
-        TERMINATED = 2
-    };
-
     using ScanResultCallback = void (*)(const NetworkCommissioning::WiFiScanResponse &);
-    using ScanDoneCallback   = void (*)(WiFiRequestStatus);
-    using ConnectionCallback = void (*)();
+    using ScanDoneCallback   = void (*)(const wifi_status &);
+    using ConnectionCallback = void (*)(int);
```
markaj-nordic, line 93: Why not using `wifi_conn_status` instead of `int`?

src/platform/nrfconnect/wifi/NrfWiFiDriver.cpp
markaj-nordic: discussed f2f, will be fixed later as it's not vital for now.

src/platform/nrfconnect/wifi/WiFiManager.h
LuDuda: Agree.. not sure, why I used `int` :)

PR #405, opened 2024-03-08
[nrf noup] Fix zap instalation for MacOs

LipinskiPNordicSemi:
Fix updating get_zap.py script, which ensure zap will work on the MacOs systems.

### DIFF
--- a/scripts/setup/nrfconnect/get_zap.py
+++ b/scripts/setup/nrfconnect/get_zap.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 #
-# Copyright (c) 2023 Project CHIP Authors
+# Copyright (c) 2024 Project CHIP Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 # limitations under the License.
 #
 
+#
 # The script parses integrations/docker/images/chip-build/Dockerfile file
 # from the Matter repository and looks for ENV ZAP_VERSION= string to find
 # currently recommended ZAP version. After that the package matching this version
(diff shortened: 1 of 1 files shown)


scripts/setup/nrfconnect/get_zap.py
```diff
@@ -198,4 +204,5 @@ def main():
 
 
 if __name__ == '__main__':
+
```
Damian-Nordic, line 207: nit but why are blank lines added in `if` or a method block but only sometimes. Let's be consistent :)

scripts/setup/nrfconnect/get_zap.py
```diff
@@ -118,19 +118,27 @@ def set_executable(location, package_name, filename):
...
 def unzip_zap_package(location, package_name):
+    package = location + f"/{package_name}.zip"
+    destination = location + "/" + package_name
+
     try:
-        zip = ZipFile(os.path.join(location, f"{package_name}.zip"))
-        zip.extractall(os.path.join(location, package_name))
-        zip.close()
+        if (platform.system() == 'Darwin'):
+            subprocess.run(['unzip', package, '-d', destination], stdout = subprocess.PIPE)
```
Damian-Nordic, line 126: Use `subprocess.check_call` or error handling? Note that using `subprocess.PIPE` without consuming the output is dangerous as it may block the child process if PIPE buffer is filled. Btw, what was the issue with `ZipFile`? Should we add a comment for the future reference?

Damian-Nordic (approved):

scripts/setup/nrfconnect/get_zap.py
LipinskiPNordicSemi: Usage of any python library for zip files on the MacOs fails at the unziping Alias files process, i mean it just unzip it as a .txt files. Thi is the reason I change it for the shell command. 
About subprocess.check_call, good point, I will add this.

scripts/setup/nrfconnect/get_zap.py
```diff
@@ -1,7 +1,7 @@
-#!/usr/bin/env python3
 
+#!/usr/bin/env python3
```
Damian-Nordic, line 2: It's better to keep the shebang line as the first one. I'm not sure if all shells can handle a blank line.

ArekBalysNordic (approved):

LuDuda (approved):

PR #407, opened 2024-03-11
[nrf noup] Fix blocker issue with stack sizes

ArekBalysNordic:
The stack sizes are too low when Oberon and CC3XX drivers are enabled.


config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -229,6 +229,7 @@ config IEEE802154_NRF5_RX_STACK_SIZE
     default 1024
 
 config OPENTHREAD_THREAD_STACK_SIZE
+    default 6144 if PSA_CRYPTO_DRIVER_CC3XX && OBERON_BACKEND
```
LuDuda, line 232:
```suggestion
    default 6144 if PSA_CRYPTO_DRIVER_CC3XX && PSA_CRYPTO_DRIVER_OBERON
```

And same for other?

LuDuda (approved):

src/platform/nrfconnect/CHIPDevicePlatformConfig.h
```diff
@@ -215,11 +215,7 @@
 #endif // CHIP_DEVICE_CONFIG_CHIP_TASK_PRIORITY
 
 #ifndef CHIP_DEVICE_CONFIG_CHIP_TASK_STACK_SIZE
-#ifdef CONFIG_LTO
```
Damian-Nordic, line 218: Why this change? It's not mentioned in the PR description.

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -229,6 +229,7 @@ config IEEE802154_NRF5_RX_STACK_SIZE
     default 1024
 
 config OPENTHREAD_THREAD_STACK_SIZE
+    default 6144 if PSA_CRYPTO_DRIVER_CC3XX && PSA_CRYPTO_DRIVER_OBERON
```
Damian-Nordic, line 232: I know we may be in hurry, but shouldn't OpenThread provide the valid defaults for this option?

Damian-Nordic (approved):

config/nrfconnect/chip-module/Kconfig.defaults
ArekBalysNordic: I don't know if they have such a case when they use both drivers at the same time. We fell into this issue when we enabled support for the `ECDSA_SECP_K1_256` curve.

ArekBalysNordic: So it is a very specific case.

src/platform/nrfconnect/CHIPDevicePlatformConfig.h
ArekBalysNordic: We need to increase the stack size to 8k anyway, so it doesn't matter whether it is the LTO build or not.

ArekBalysNordic: But... It is regarding CC + Oberon only, so I will add the ifdef there to not impact on the memory requirements table,

ArekBalysNordic:
The unit test fails are not related to this PR.

markaj-nordic (approved):

Damian-Nordic (approved):

PR #412, opened 2024-03-14
[nrf noup] Wifi directed scan with single scan

LipinskiPNordicSemi:
This commit implements directed scanning and single ssid scanning while commissioning.

### DIFF
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -188,7 +188,22 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
     mWiFiState          = WIFI_STATE_SCANNING;
     mSsidFound          = false;
 
-    if (0 != net_mgmt(NET_REQUEST_WIFI_SCAN, mNetIf, NULL, 0))
+    wifi_scan_params * scanParams{ nullptr };
+    size_t scanParamsSize{ 0 };
+
+    if (!ssid.empty())
+    {
+        /* We must assume that the ssid is handled as a NULL-terminated string.
+           Note that the mScanSsidBuffer is initialized with zeros. */
+        VerifyOrReturnError(ssid.size() < sizeof(mScanParams), CHIP_ERROR_INVALID_ARGUMENT);
+        memcpy(mScanSsidBuffer, ssid.data(), ssid.size());
+        mScanSsidBuffer[ssid.size()] = 0; // indicate the end of ssid string
+        mScanParams.ssids[0]         = mScanSsidBuffer;
+        mScanParams.ssids[1]         = nullptr; // indicate the end of ssids list
+        scanParams                   = &mScanParams;
+        scanParamsSize               = sizeof(*scanParams);
+    }
+    if (0 != net_mgmt(NET_REQUEST_WIFI_SCAN, mNetIf, scanParams, scanParamsSize))
     {
         ChipLogError(DeviceLayer, "Scan request failed");
         return CHIP_ERROR_INTERNAL;
--- a/src/platform/nrfconnect/wifi/WiFiManager.h
+++ b/src/platform/nrfconnect/wifi/WiFiManager.h
@@ -220,7 +220,9 @@ private:
 
     net_if * mNetIf{ nullptr };
     ConnectionParams mWiFiParams{};
-    ConnectionHandling mHandling;
+    ConnectionHandling mHandling{};
+    wifi_scan_params mScanParams{};
+    char mScanSsidBuffer[DeviceLayer::Internal::kMaxWiFiSSIDLength + 1] = { 0 };
     wifi_iface_state mWiFiState;
     wifi_iface_state mCachedWiFiState;
     net_mgmt_event_callback mWiFiMgmtClbk{};


src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -188,14 +188,26 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
     mWiFiState          = WIFI_STATE_SCANNING;
     mSsidFound          = false;
 
-    if (0 != net_mgmt(NET_REQUEST_WIFI_SCAN, mNetIf, NULL, 0))
+    if (!ssid.empty())
     {
-        ChipLogError(DeviceLayer, "Scan request failed");
-        return CHIP_ERROR_INTERNAL;
+        memcpy(mScanSsidBuffer, ssid.data(), ssid.size());
+        mScanParams.ssids[0] = reinterpret_cast<const char *>(&mScanSsidBuffer);
```
Damian-Nordic, line 194: I think SSID and scan params could be allocated on stack. Also, make sure you allocate enough buffer for null-terminator and validate the input string size:

```suggestion
        char ssidBuffer[DeviceLayer::Internal::kMaxWiFiSSIDLength + 1] = {};
        wifi_scan_params scanParams = {};
        
        VerifyOrReturnError(ssid.size() < sizeof(ssidBuffer), CHIP_ERROR_INVALID_ARGUMENT);
        memcpy(ssidBuffer, ssid.data(), ssid.size());
        scanParams.ssid[0] = ssidBuffer;
```

Damian-Nordic (approved):
I think this is a better approach than filtering results in callback.

src/platform/nrfconnect/wifi/WiFiManager.cpp
LipinskiPNordicSemi: The problem with those variables on the stac is that I think the function is finished before scanning (scanning might be done on the zephyr thead).
If I try directed scanning with your suggestions, the result is like without SSID given.

LipinskiPNordicSemi: After implementing suggested changes, I found some problems with directed scanning - one in few scan couldn't find network by ssid

Damian-Nordic: ok, nevermind then but please validate the SSID length in your solution and add `+1` to the buffer size :)

LuDuda (approved):
Looks great!

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Looks like the Wi-Fi driver uses only the shallow copy of `wifi_scan_params`, so we cannot rely on stack allocated objects here.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -188,7 +188,18 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
...
 
-    if (0 != net_mgmt(NET_REQUEST_WIFI_SCAN, mNetIf, NULL, 0))
+    wifi_scan_params * scanParamsFinal{ nullptr };
+
+    if (!ssid.empty())
+    {
+        /* We must assume that the ssid is handled as a NULL-terminated string.
+           Note that the mScanSsidBuffer is initialized with zeros. */
+        VerifyOrReturnError(ssid.size() < sizeof(mScanParams), CHIP_ERROR_INVALID_ARGUMENT);
+        memcpy(mScanSsidBuffer, ssid.data(), ssid.size());
```
Damian-Nordic, line 198: Can you add null-terminator here? Thinking about the case in which we first scan for "my-network-1" and then some shorter "my-network".

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -188,7 +188,18 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
...
+        mScanParams.ssids[0] = mScanSsidBuffer;
+        scanParamsFinal      = &mScanParams;
+    }
+    if (0 != net_mgmt(NET_REQUEST_WIFI_SCAN, mNetIf, scanParamsFinal, sizeof(*scanParamsFinal)))
```
Damian-Nordic, line 202: Is it OK to pass non-zero size even if params are null?

Damian-Nordic (approved):

src/platform/nrfconnect/wifi/WiFiManager.cpp
Damian-Nordic: Thanks for checking!

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -188,7 +188,18 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
...
-    if (0 != net_mgmt(NET_REQUEST_WIFI_SCAN, mNetIf, NULL, 0))
+    wifi_scan_params * scanParamsFinal{ nullptr };
+
+    if (!ssid.empty())
+    {
+        /* We must assume that the ssid is handled as a NULL-terminated string.
+           Note that the mScanSsidBuffer is initialized with zeros. */
+        VerifyOrReturnError(ssid.size() < sizeof(mScanParams), CHIP_ERROR_INVALID_ARGUMENT);
+        memcpy(mScanSsidBuffer, ssid.data(), ssid.size());
+        mScanParams.ssids[0] = mScanSsidBuffer;
```
krish2718, line 199: as there is no length of SSIDs, the driver relies on `ssids[1] = NULL` to know the length.

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: I think so. But, ok, let's be more cautious.

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: I see, shall we set the max_bss_cnt to 1 in such a case?

src/platform/nrfconnect/wifi/WiFiManager.cpp
markaj-nordic: Valid point, thanks.

src/platform/nrfconnect/wifi/WiFiManager.cpp
krish2718: no, `max_bss_cnt` is for limiting scan results, it's not used for limiting SSIDs

markaj-nordic:
@krish2718, @Damian-Nordic - please check one last time.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -188,7 +188,22 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
     mWiFiState          = WIFI_STATE_SCANNING;
     mSsidFound          = false;
 
-    if (0 != net_mgmt(NET_REQUEST_WIFI_SCAN, mNetIf, NULL, 0))
+    wifi_scan_params * scanParamsFinal{ nullptr };
+    size_t scanParamsLen{ 0 };
```
LuDuda, line 192:
```suggestion
    size_t scanParamsSize{ 0 };
```

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -188,7 +188,22 @@ CHIP_ERROR WiFiManager::Scan(const ByteSpan & ssid, ScanResultCallback resultCal
     mWiFiState          = WIFI_STATE_SCANNING;
     mSsidFound          = false;
 
-    if (0 != net_mgmt(NET_REQUEST_WIFI_SCAN, mNetIf, NULL, 0))
+    wifi_scan_params * scanParamsFinal{ nullptr };
```
LuDuda, line 191:
```suggestion
    wifi_scan_params * scanParams{ nullptr };
```

LuDuda (approved):

Damian-Nordic (approved):
Looks great now

PR #418, opened 2024-04-03
[nrf fromlist][nrfconnect] Improve CHIP_MEMORY_PROFILING config

ArekBalysNordic:
Enabled the full SHELL option when memory profiling is enabled.

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig.defaults
+++ b/config/nrfconnect/chip-module/Kconfig.defaults
@@ -465,7 +465,7 @@ config SHELL_STACK_SIZE
     default 2616 if CHIP_WIFI
 
 config SHELL_MINIMAL
-    default y
+    default y if !CHIP_MEMORY_PROFILING
 
 config KERNEL_SHELL
     default n
--- a/config/nrfconnect/chip-module/Kconfig.features
+++ b/config/nrfconnect/chip-module/Kconfig.features
@@ -75,15 +75,20 @@ endif # CHIP_SPI_NOR
 
 config CHIP_MEMORY_PROFILING
 	bool "Enable features for tracking memory usage"
+	# Matter stack
 	select CHIP_STATISTICS
+	# Heap
 	select CHIP_MALLOC_SYS_HEAP_WATERMARKS_SUPPORT if CHIP_MALLOC_SYS_HEAP
+	select SYS_HEAP_RUNTIME_STATS if CHIP_MALLOC_SYS_HEAP	
+	# Crypto
 	select MBEDTLS_MEMORY_DEBUG if !CHIP_CRYPTO_PSA
-	select SYS_HEAP_RUNTIME_STATS if CHIP_MALLOC_SYS_HEAP
-	select KERNEL_SHELL
+	# Network
 	select NET_STATISTICS
 	select NET_SHELL
 	select NET_BUF_POOL_USAGE
 	select OPENTHREAD_SHELL if !CHIP_WIFI
+	# Zephyr
+	select KERNEL_SHELL
 	help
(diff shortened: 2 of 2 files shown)


config/nrfconnect/chip-module/Kconfig.features
```diff
@@ -75,18 +75,31 @@ endif # CHIP_SPI_NOR
...
+	# Matter stack
 	select CHIP_STATISTICS
+	# Heap
 	select CHIP_MALLOC_SYS_HEAP_WATERMARKS_SUPPORT if CHIP_MALLOC_SYS_HEAP
-	select MBEDTLS_MEMORY_DEBUG if !CHIP_CRYPTO_PSA
-	select SYS_HEAP_RUNTIME_STATS if CHIP_MALLOC_SYS_HEAP
-	select KERNEL_SHELL
+	select SYS_HEAP_RUNTIME_STATS if CHIP_MALLOC_SYS_HEAP	
+	# Crypto
+	select MBEDTLS_MEMORY_DEBUG if CHIP_CRYPTO_MBEDTLS
```
Damian-Nordic, line 84: We don't define `CHIP_CRYPTO_MBEDTLS` Kconfig, do we?

config/nrfconnect/chip-module/Kconfig.features
```diff
@@ -75,18 +75,31 @@ endif # CHIP_SPI_NOR
...
 	select OPENTHREAD_SHELL if !CHIP_WIFI
+	# Zephyr
+	select KERNEL_SHELL
 	help
 	  Enables features for tracking memory usage in Matter.
 
+# Enable full debug option when memory profiling is set.
+if CHIP_MEMORY_PROFILING
+
+config SHELL_MINIMAL
```
Damian-Nordic, line 98: We have a similar default in `Kconfig.defaults`. Which one is selected depends on whether a sample includes `Kconfig.defaults` or `Kconfig.features` first (and it seems we're not consistent wrt that). I suggest modifying `Kconfig.defaults` with proper:
```
default y if !CHIP_MEMORY_PROFILING
```
to not rely on the order of inclusion.

Damian-Nordic (approved):

Damian-Nordic:
Also, can you please describe why SHELL_MINIMAL is undesired? It doesn't seem related to the profiling feature.

ArekBalysNordic:
> Also, can you please describe why SHELL_MINIMAL is undesired? It doesn't seem related to the profiling feature.

It was requested by QA to have it because it helps with using all SHELL components like KERNEL_SHELL, CHIP_SHELL, and NET_SHELL.

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -465,10 +465,10 @@ config SHELL_STACK_SIZE
     default 2616 if CHIP_WIFI
 
 config SHELL_MINIMAL
-    default y
+    default y if !CHIP_MEMORY_PROFILING
 
 config KERNEL_SHELL
-    default n
+    default n if !CHIP_MEMORY_PROFILING
```
LuDuda, line 471: Do we need it? Could we just enable KERNEL_SHELL inside CHIP_MEMORY_PROFILING Kconfig definition?

ArekBalysNordic: I did it to ensure that the order of kconfig includes doesn't matter as Damian noticed above.

ArekBalysNordic: So indeed, it is not needed then 😄 I thought that it also will be overwritten.

kkasperczyk-no (approved):

PR #419, opened 2024-04-05
[scripts] Fixed nrfconnect factory generation scripts

kkasperczyk-no:
The factory_data.hex file is not generated as an intermediate product of factory generation process. In result, factory generation works only if merging with firmware is used or scripts are manually invoked.

Added optional --size and --offset arguments to
generate_nrfconnect_chip_factory_data.py script that results in calling nrfconnect_generate_partition.py internally. It solves an issue and additionally simplifies manual generation process (if selected).

### DIFF
--- a/config/nrfconnect/chip-module/generate_factory_data.cmake
+++ b/config/nrfconnect/chip-module/generate_factory_data.cmake
@@ -15,25 +15,29 @@
 #
 
 
-# Create a JSON file based on factory data given via kConfigs.
+# Create a .hex file in CBOR format based on factory data given via kConfigs.
 #
 # This function creates a list of arguments for external script and then run it to write a JSON file.
 # Created JSON file can be checked using JSON SCHEMA file if it is provided.
+# Next, the resulting .hex file is generated based on previously created JSON file.
 #
 # This script can be manipulated using following kConfigs:
 # - To merge generated factory data with final zephyr.hex file set kConfig CONFIG_CHIP_FACTORY_DATA_MERGE_WITH_FIRMWARE=y
 # - To use default certification paths set CONFIG_CHIP_FACTORY_DATA_USE_DEFAULTS_CERTS_PATH=y 
 # 
-# During generation process a some file will be created in zephyr's build directory:
+# During generation process the following files will be created in zephyr's build directory:
 # - <factory_data_target>.json a file containing all factory data written in JSON format.
+# - <factory_data_target>.hex a file containing all factory data in CBOR format.
+# - <factory_data_target>.bin a binary file containing all raw factory data in CBOR format.
+# - <factory_data_target>.cbor a file containing all factory data in CBOR format.
 #
 # [Args]:
 #   factory_data_target - a name for target to generate factory_data.
 #   script_path         - a path to script that makes a JSON factory data file from given arguments.
 #   schema_path         - a path to JSON schema file which can be used to verify generated factory data JSON file.
 #                         This argument is optional, if you don't want to verify the JSON file put it empty "".
-#   output_path         - a path to output directory, where created JSON file will be stored.
-function(nrfconnect_create_factory_data_json factory_data_target script_path schema_path output_path)
+#   output_path         - a path to output directory, where created hex and JSON files will be stored.
+function(nrfconnect_create_factory_data factory_data_target script_path schema_path output_path)
 
 # set script args for future purpose
 set(script_args)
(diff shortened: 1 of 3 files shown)


scripts/tools/nrfconnect/generate_nrfconnect_chip_factory_data.py
```diff
@@ -244,8 +246,8 @@ def _validate_args(self):
             "Cannot find Spake2+ verifier, to generate a new one please provide passcode (--passcode)"
         assert ((self._args.gen_certs and self._args.chip_cert_path) or (self._args.dac_cert and self._args.pai_cert and self._args.dac_key)), \
             "Cannot find paths to DAC or PAI certificates .der files. To generate a new ones please provide a path to chip-cert executable (--chip_cert_path) and add --gen_certs argument"
-        assert self._args.output.endswith(".json"), \
-            "Output path doesn't contain .json file path. ({})".format(self._args.output)
+        assert not (self._args.output.endswith(".json")), \
+            "Output path contain .json extension in path that is redundant. ({})".format(self._args.output)
```
ArekBalysNordic, line 250: If we want to call assert on that, this should point out that the .json extension in the path is *forbidden* instead of *redundant*. 
Apart from that, I think that acutally it is redundant, but the script should remove the extension to keep backward compatibility.

scripts/tools/nrfconnect/generate_nrfconnect_chip_factory_data.py
```diff
@@ -551,9 +559,9 @@ def base64_str(s): return base64.b64decode(s)
         log.basicConfig(format='[%(levelname)s] %(message)s', level=log.INFO)
 
     # check if json file already exist
-    if (exists(args.output) and not args.overwrite):
+    if (exists(args.output+".json") and not args.overwrite):
```
ArekBalysNordic, line 562:
```suggestion
    if (exists(args.output + ".json") and not args.overwrite):
```

ArekBalysNordic (commented):
Please notice, that the error in Github Action is relevant to this PR. :)

scripts/tools/nrfconnect/generate_nrfconnect_chip_factory_data.py
kkasperczyk-no: It is mentioned that the .json should not be provided in the argument description:

The .json extension will be automatically added by the script and should not be provided.

Do you mean we should support both versions - with and without extension?

config/nrfconnect/chip-module/generate_factory_data.cmake
```diff
@@ -15,25 +15,29 @@
 #
 
 
-# Create a JSON file based on factory data given via kConfigs.
+# Create a .hex file in CBOR format based on factory data given via kConfigs.
```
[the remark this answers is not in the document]
kkasperczyk-no: Yes, I know, but is it important from the cmake perspective? I didn't write about it, because after removing this split on two functions, we create .hex with this one and .json is used only internally as an intermediate product.

It looks a little bit different for the manual generation, the it's possible to generate only .json, but in cmake it will not be possible anymore.

scripts/tools/nrfconnect/generate_nrfconnect_chip_factory_data.py
ArekBalysNordic: Yes, I meant to support both ways because in the other case, we break compatibility with the previous versions of the scripts. I guess it is a one-liner but it will help customers that may want to use this script as previously.

kkasperczyk-no: Ok

config/nrfconnect/chip-module/generate_factory_data.cmake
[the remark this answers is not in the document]
ArekBalysNordic: It is minor because I guess only a few people look at this description, but we don't remove that intermediate product and somebody may want to use it somehow.

ArekBalysNordic (approved):

config/nrfconnect/chip-module/generate_factory_data.cmake
```diff
@@ -15,25 +15,29 @@
...
 #
 # This function creates a list of arguments for external script and then run it to write a JSON file.
 # Created JSON file can be checked using JSON SCHEMA file if it is provided.
+# Next the result .hex file is generated based on previously created JSON file.
```
markaj-nordic, line 22:
```suggestion
# Next, the resulting .hex file is generated based on previously created JSON file.
```

config/nrfconnect/chip-module/generate_factory_data.cmake
```diff
@@ -15,25 +15,29 @@
...
 #
 # This script can be manipulated using following kConfigs:
 # - To merge generated factory data with final zephyr.hex file set kConfig CONFIG_CHIP_FACTORY_DATA_MERGE_WITH_FIRMWARE=y
 # - To use default certification paths set CONFIG_CHIP_FACTORY_DATA_USE_DEFAULTS_CERTS_PATH=y 
 # 
 # During generation process a some file will be created in zephyr's build directory:
```
markaj-nordic, line 28:
```suggestion
# During generation process the following files will be created in zephyr's build directory:
```

config/nrfconnect/chip-module/generate_factory_data.cmake
```diff
@@ -120,62 +124,29 @@ if(CONFIG_CHIP_DEVICE_ENABLE_KEY)
...
 
-# Set output JSON file and path to SCHEMA file to validate generated factory data
-set(factory_data_json ${output_path}/${factory_data_target}.json)
-string(APPEND script_args "-o \"${factory_data_json}\"\n")
+# Set output path and path to SCHEMA file to validate generated factory data
+set(factory_data_output_path ${output_path}/${factory_data_target})
+string(APPEND script_args "-o \"${factory_data_output_path}\"\n")
 string(APPEND script_args "-s \"${schema_path}\"\n")
 
+# Add optional offset and size arguments to generate .hex file as well as .json.
```
markaj-nordic, line 132:
```suggestion
# Add optional offset and size arguments to generate both .hex and .json file.
```

docs/guides/nrfconnect_factory_data_configuration.md
```diff
@@ -218,14 +218,19 @@ file written in another way. To make sure that the JSON file is correct and the
 device is able to read out parameters,
 [verify the file using the JSON schema tool](#verifying-using-the-json-schema-tool).
 
-### Creating the factory data JSON file with the first script
+You can also use only the first script to generate both JSON and HEX files, by providing
+optional `offset` and `size` arguments, which results in invoking the script internally.
+Such option is recommended one, but invoking two scripts one by one is also supported
```
markaj-nordic, line 223:
```suggestion
Such option is the recommended one, but invoking two scripts one by one is also supported
```

docs/guides/nrfconnect_factory_data_configuration.md
```diff
@@ -341,6 +346,26 @@ To use this script, complete the following steps:
...
     > it.
 
+    k. (optional) Partition offset that is an address in device's NVM memory, where factory data will be stored.
+
+    ```
+    --offset <offset>
+    ```
+
+    > **Note:** To generate a HEX file with factory data, you need to provide both `offset` and `size` optional arguments.
+    > As a result, `factory_data.hex` and `factory_data.bin` files are created in the `output` directory. The first file contains the memory offset.
```
markaj-nordic, line 356:
```suggestion
    > As a result, `factory_data.hex` and `factory_data.bin` files are created in the `output` directory. The first file contains the required memory offset.
```

docs/guides/nrfconnect_factory_data_configuration.md
```diff
@@ -341,6 +346,26 @@ To use this script, complete the following steps:
...
+    > For this reason, it can be programmed directly to the device using a programmer (for example, `nrfjprog`).
+
+    l. (optional) The maximum partition size in device's NVM memory, where factory data will be stored.
+
+    ```
+    --size <size>
+    ```
+
+    > **Note:** To generate a HEX file with factory data, you need to provide both `offset` and `size` optional arguments.
+    > As a result, `factory_data.hex` and `factory_data.bin` files are created in the `output` directory. The first file contains the memory offset.
```
markaj-nordic, line 366: ditto

scripts/tools/nrfconnect/generate_nrfconnect_chip_factory_data.py
```diff
@@ -310,9 +314,9 @@ def generate_json(self):
             sys.exit(-1)
 
         try:
-            json_file = open(self._args.output, "w+")
+            json_file = open(self._args.output+".json", "w+")
```
markaj-nordic, line 317: For clarity we may keep the `self._args.output+".json` as a constant named e.g. `output`/`final_output`.

markaj-nordic (approved):
LGTM. Only some minors found.

PR #425, opened 2024-04-09
[nrf noup] Change of log configs with optimization

HubertHugo:
Many logs were not handled when CONFIG_CHIP_LOG_SIZE_OPTIMIZATION = 1. This PR makes clear which logs are used and which are blocked. More log modules are added to blocked when optimization is on.


ArekBalysNordic:
Seems not needed anymore.

PR #426, opened 2024-04-10
[nrf toup] nrfconnect: enable detailed logging of Matter messages

LuDuda:
This commit adds an option to enable IM Pretty Print from the KConfig.

(to be upstreamed first -> then I will change commit to nrf fromlist)


Damian-Nordic (approved):

LuDuda:
Will be a part of larger PR later: https://github.com/doublemis1/connectedhomeip/pull/129

doublemis1:
If we want to make it work instantly we should enable DataMangment Detail Logging:
https://github.com/nrfconnect/sdk-connectedhomeip/blob/master/src/platform/nrfconnect/CHIPPlatformConfig.h#L117 (which also enable additional logs) or
change the log level for pretty print probably here:
https://github.com/nrfconnect/sdk-connectedhomeip/blob/579ef888336b3e83f8a5dbd518911accbcb21d11/src/app/MessageDef/MessageDefHelper.cpp#L50C45-L50C56

We can also think about enabling CHIP_DETAIL_LOGGING if is not enable:
https://github.com/nrfconnect/sdk-connectedhomeip/blob/579ef888336b3e83f8a5dbd518911accbcb21d11/src/app/MessageDef/StatusIB.cpp#L87C5-L87C24

LuDuda:
Closed stale PR. Part of it already contributed.

PR #427, opened 2024-04-17
[nrf toup] [nrfconnect] Disable External FD if QSPI NOR is not available

ArekBalysNordic:
We need to have support for factory data when the partition manager is enabled but there is no support for QSPI external flash.

### DIFF
--- a/src/platform/nrfconnect/FactoryDataProvider.cpp
+++ b/src/platform/nrfconnect/FactoryDataProvider.cpp
@@ -509,9 +509,10 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::GetUserKey(const char * userKe
 
 // Fully instantiate the template class in whatever compilation unit includes this file.
 template class FactoryDataProvider<InternalFlashFactoryData>;
-#if defined(USE_PARTITION_MANAGER) && USE_PARTITION_MANAGER == 1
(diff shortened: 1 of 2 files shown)


src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -509,7 +509,7 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::GetUserKey(const char * userKe
 
 // Fully instantiate the template class in whatever compilation unit includes this file.
 template class FactoryDataProvider<InternalFlashFactoryData>;
-#if defined(USE_PARTITION_MANAGER) && USE_PARTITION_MANAGER == 1
+#if defined(USE_PARTITION_MANAGER) && USE_PARTITION_MANAGER == 1 && defined(CONFIG_CHIP_QSPI_NOR)
```
kkasperczyk-no, line 512: And what about external flash over SPI?

kkasperczyk-no (commented):
Overall please remember about contributing this one and previous factory data related nrf toup commits to upstream.

ArekBalysNordic:
> The `ExternalFlashFactoryData` is not used anywhere anyway, so was there a problem with that or it's just a matter of optimization?
> 
> Should we add the same condition when instantiating the concrete template classes in at the bottom of the FactoryDataProvider.cpp file?

Actually, I don't know why we haven't removed the external flash support from there yet...

kkasperczyk-no (approved):

PR #428, opened 2024-04-18
[mrp] Increase default retry interval for Thread

Damian-Nordic:
The current 800ms is not enough in real setups, where Thread routers must serve as intermediate hops for many parallel conversations. Bump this to 2s.


src/platform/nrfconnect/CHIPPlatformConfig.h
```diff
@@ -106,21 +106,16 @@
 #define CHIP_CONFIG_LOG_MODULE_Support_PROGRESS 0
 #endif
 
-// Set MRP retry intervals for Thread and Wi-Fi to test-proven values.
 #ifndef CHIP_CONFIG_MRP_LOCAL_ACTIVE_RETRY_INTERVAL
-#if CHIP_ENABLE_OPENTHREAD
-#define CHIP_CONFIG_MRP_LOCAL_ACTIVE_RETRY_INTERVAL (800_ms32)
-#else
+#ifndef CONFIG_NET_L2_OPENTHREAD
```
LuDuda, line 110: Do we need it, or should leave default values for Wi-Fi as well?

LuDuda (approved):

src/platform/nrfconnect/CHIPPlatformConfig.h
Damian-Nordic: You mean 300ms? I thought we had a reason for increasing it on Wi-Fi so didn't want to change it.

LuDuda: Oh, i thought that we set 2s for both Thread and Wifi by default, and here we set 1s for Wi-Fi.. But this is not the case.

LuDuda (approved):

kkasperczyk-no (approved):

PR #433, opened 2024-04-24
[nrf noup] Workaround for PSA P256Keypair Generate

ArekBalysNordic:
This is a workaround to ensure thread-safety for P256Keypair generation when both Matter and Openthread threads want to generate keys.

### DIFF
--- a/src/crypto/PSAOperationalKeystore.cpp
+++ b/src/crypto/PSAOperationalKeystore.cpp
@@ -22,6 +22,10 @@
 
 #include <psa/crypto.h>
 
+// TODO: This is a temporary workaround until the PSA Core is not thread-safe.
+// The fix is going to be provided in MbedTLS 3.6.0
+#include <platform/CHIPDeviceLayer.h> // nogncheck
+
 namespace chip {
 namespace Crypto {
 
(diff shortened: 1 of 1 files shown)


src/crypto/PSAOperationalKeystore.cpp
```diff
@@ -63,6 +65,10 @@ CHIP_ERROR PSAOperationalKeystore::PersistentP256Keypair::Generate()
 
     Destroy();
 
+#ifdef CONFIG_NET_L2_OPENTHREAD
+    chip::DeviceLayer::ThreadStackMgr().LockThreadStack();
```
Damian-Nordic, line 69: Worth adding a comment that it's a temporary hack.

Damian-Nordic (approved):

src/crypto/PSAOperationalKeystore.cpp
ArekBalysNordic: Added :)

markaj-nordic (approved):

ArekBalysNordic:
Closing, because it will be repaired in the Crypto core.

PR #449, opened 2024-05-29
[nrf toup][nrfconnect] Removed defaults for child images

kkasperczyk-no:
Due to the sysbuild integration and child images deprecation the default configuration for child images has been removed.


ArekBalysNordic (approved):
I guess [nrf toup] will not work for this separate commit since samples in upstream are not ready for that :) Maybe we should point it out somehow?

kkasperczyk-no:
> I guess [nrf toup] will not work for this separate commit since samples in upstream are not ready for that :) Maybe we should point it out somehow?

Yeah I know. I didn't want to modify the child_images content, because probably we will contribute this commit with the NCS 2.7.0 and start using sysbuild in the upstream, so it's pointless. I thought we will introduce sysbuild support to the examples with the next NCS upmerge and this commit will be part of it. What do you think about it?

ArekBalysNordic:
> > I guess [nrf toup] will not work for this separate commit since samples in upstream are not ready for that :) Maybe we should point it out somehow?
> 
> Yeah I know. I didn't want to modify the child_images content, because probably we will contribute this commit with the NCS 2.7.0 and start using sysbuild in the upstream, so it's pointless. I thought we will introduce sysbuild support to the examples with the next NCS upmerge and this commit will be part of it. What do you think about it?

OK, let's do it in this way :)

LuDuda (approved):

PR #450, opened 2024-06-04
[nrf noup] Power nRF54H20 workaround.

ArekBalysNordic:
Provided a workaround for power reason for nrf54h20 since it is not available yet.

### DIFF
--- a/src/platform/nrfconnect/Reboot.cpp
+++ b/src/platform/nrfconnect/Reboot.cpp
@@ -21,14 +21,14 @@
 
 #include <zephyr/sys/reboot.h>
 
-#ifndef CONFIG_ARCH_POSIX
+#if !(defined(CONFIG_ARCH_POSIX) || defined(CONFIG_SOC_SERIES_NRF54HX))
 #include <hal/nrf_power.h>
 #endif
 
 namespace chip {
 namespace DeviceLayer {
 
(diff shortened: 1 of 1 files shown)


src/platform/nrfconnect/Reboot.cpp
```diff
@@ -21,14 +21,14 @@
...
-#ifndef CONFIG_ARCH_POSIX
+#if !(defined(CONFIG_ARCH_POSIX) || defined(CONFIG_BOARD_NRF54H20DK_NRF54H20_CPUAPP))
 #include <hal/nrf_power.h>
 #endif
 
 namespace chip {
 namespace DeviceLayer {
 
-#ifdef CONFIG_ARCH_POSIX
+#if defined(CONFIG_ARCH_POSIX) || defined(CONFIG_BOARD_NRF54H20DK_NRF54H20_CPUAPP)
```
kkasperczyk-no, line 31: TBH this is quite ugly workaround :) Could we at least use the SoC name and not the DK in the source code? This will not apply if someone would create own device based on our SoC.

ArekBalysNordic: Right, I will change it.

kkasperczyk-no (approved):

PR #452, opened 2024-06-06
[nrf toup][nrfconnect] Fixed paths to used in the guides

kkasperczyk-no:
Guides contain path examples referring to the older build system version and they have to be updated.


ArekBalysNordic:
It has already been merged with another PR.

PR #454, opened 2024-06-17
[nrf noup] Fix "west zap-gui" on Windows

Damian-Nordic:
Use zap-cli.exe instead of zap.exe to determine the current ZAP version. The latter seems to generate unexpected output on Windows.


ArekBalysNordic (approved):

LuDuda (approved):
However.. this does not seem to be an issue on my set-up. Wondering if we should proceed with it to NCS 2.7.0 release - @Damian-Nordic ?

PR #457, opened 2024-06-24
[nrf toup][nrfconnect] Introduced Kconfigs to configure few params

kkasperczyk-no:
Introduced new Kconfig options to be able to configure the parameters related to persistent subscriptions re-establishment.

manifest-pr-skip


src/platform/nrfconnect/CHIPPlatformConfig.h
```diff
@@ -78,8 +78,34 @@
...
+#ifndef CONFIG_CHIP_MAX_ACTIVE_CASE_CLIENTS
+#define CHIP_CONFIG_DEVICE_MAX_ACTIVE_CASE_CLIENTS CONFIG_CHIP_MAX_ACTIVE_CASE_CLIENTS
+#endif // CONFIG_CHIP_MAX_ACTIVE_CASE_CLIENTS
+#endif // CHIP_CONFIG_DEVICE_MAX_ACTIVE_CASE_CLIENTS
+
+#ifndef CHIP_CONFIG_DEVICE_MAX_ACTIVE_DEVICES
+#ifndef CONFIG_CHIP_MAX_ACTIVE_DEVICES
+#define CHIP_CONFIG_DEVICE_MAX_ACTIVE_DEVICES CONFIG_CHIP_MAX_ACTIVE_DEVICES
+#endif // CONFIG_CHIP_MAX_ACTIVE_DEVICES
+#endif // CHIP_CONFIG_DEVICE_MAX_ACTIVE_DEVICES
```
Damian-Nordic, line 96:
```suggestion
#define CHIP_CONFIG_DEVICE_MAX_ACTIVE_DEVICES CONFIG_CHIP_MAX_ACTIVE_DEVICES
```
`#ifndef CONFIG_` should be `#ifdef CONFIG_`.

But anyway, this Kconfig should always be defined, right? And if it's defined, it's better to always use it. Otherwise, someone may not notice that these Kconfigs are set but are ineffective because of some forgotten define in `chip_project_config.h`.

kkasperczyk-no: > #ifndef CONFIG_ should be #ifdef CONFIG_.

Oh, of course, thanks!

> And if it's defined, it's better to always use it

Do you mean that we should avoid ifndef preprocessor check for defines? I actually just followed the example of already existing Kconfigs that are mapped in the same way. But what you wrote make sense to me.

Damian-Nordic: > Do you mean that we should avoid ifndef preprocessor check for defines? I actually just followed the example of already existing Kconfigs that are mapped in the same way. But what you wrote make sense to me.

Yes, I know we had a different pattern in the past but I actually removed `ifndef` from our `CHIPDevicePlatformConfig.h` after somebody was surprised that changing the discriminator in Kconfig didn't _actually_ work. I think if Kconfig allow to set all possible values, then it's OK to force using it instead of `chip_project_config.h`.

Damian-Nordic (approved):

src/platform/nrfconnect/CHIPPlatformConfig.h
kkasperczyk-no: Sure, good to know, I will follow to be consistent.

ArekBalysNordic (approved):

markaj-nordic (approved):

PR #460, opened 2024-07-09
[nrf toup][nrfconnect] Added module to synchronize DFU processes

kkasperczyk-no:
Introduced a module that can be used to synchronize DFU processes and make it mutual exclusive (prevent concurrent DFU runs using different protocols, e.g Matter OTA and DFU over BT SMP).

manifest-pr-skip


src/platform/nrfconnect/DFUSync.cpp
```diff
@@ -0,0 +1,34 @@
...
+        if (id == mOwnerId)
+        {
+            return CHIP_NO_ERROR;
+        }
+        return CHIP_ERROR_BUSY;
+    }
+
+    /* Generate random owner id to make this allocation unique. */
+    mIsTaken = true;
+    mOwnerId = chip::Crypto::GetRandU32();
```
markaj-nordic, line 17: Have you considered introducing some limits on the number of supported users of this module? By doing so we could get rid of the random generator usage and simply assign some predefined values, or monotonic counter. Asking especially because of the fact that we have only 2 DFUs mechanisms implemented.

src/platform/nrfconnect/DFUSync.h
```diff
@@ -0,0 +1,59 @@
...
+     */
+    static inline DFUSync & GetInstance()
+    {
+        static DFUSync sInstance;
+        return sInstance;
+    }
+
+private:
+	/* Mutex to synchronize the DFU operations. */
+    bool mIsTaken = false;
```
markaj-nordic, line 57: Regardless of the fact that it's protected by the `mOwnerId`, have you considered using `std::atomic<bool>` here? I guess no type is atomic by nature, even `bool`.

src/platform/nrfconnect/DFUSync.cpp
kkasperczyk-no: I didn't want to use predefined values, as this module theoretically could be used for any type of dfu in the future and maybe it should not care about it. Yeah, we could use monotonic counter, I don't know if this will be a big time optimization, but ok

src/platform/nrfconnect/DFUSync.h
kkasperczyk-no: Good idea, applied.

src/platform/nrfconnect/DFUSync.cpp
kkasperczyk-no: Replaced rand with counter.

src/platform/nrfconnect/DFUSync.cpp
```diff
@@ -0,0 +1,33 @@
...
+        if (id == mOwnerId)
+        {
+            return CHIP_NO_ERROR;
+        }
+        return CHIP_ERROR_BUSY;
+    }
+
+    mIsTaken = true;
+    /* Increment owner id to make sure that every allocation is unique. */
+    mOwnerId++;
```
markaj-nordic, line 16: Maybe we could just add some description in the pubic API that uint32_t is used to store unique ids to explicitly point out the obvious limitations.

src/platform/nrfconnect/DFUSync.h
```diff
@@ -0,0 +1,60 @@
...
+     */
+    static inline DFUSync & GetInstance()
+    {
+        static DFUSync sInstance;
+        return sInstance;
+    }
+
+private:
+	/* Mutex to synchronize the DFU operations. */
+    std::atomic<bool> mIsTaken = false;
```
markaj-nordic, line 58: Hmm I thought the copy c'tor is deleted for atomic template, but maybe I'm wrong. In case it does not compile, the following should work:

```suggestion
    std::atomic<bool> mIsTaken {false};
```

markaj-nordic (approved):

src/platform/nrfconnect/DFUSync.cpp
kkasperczyk-no: Discussed f2f

PR #461, opened 2024-07-12
[nrf toup][nrfconnect] Added support for Matter OTA using SUIT

kkasperczyk-no:
Modified the OTAImageProcessorImpl to use dfu target multi image only in case of using the mcuboot bootloader and to use dfu target in case of using suit.


ArekBalysNordic (approved):
Looks good, but maybe let's wait for the final decision, on whether we will use dfu_target or raw SUIT API.

PR #462, opened 2024-07-15
Fix ICD LIT connected configurations and `zap-generate` `-o` path argument

maciejbaczmanski:
- Fixed implying instead of selecting CIP and UAT features under `CHIP_ICD_LIT_SUPPORT`
- Fixed `-o` path argument to be relative to `.zap` file's location instead of Matter library's location.

manifest-pr-skip


ArekBalysNordic (approved):

scripts/west/zap_generate.py
```diff
@@ -49,7 +49,7 @@ def do_run(self, args, unknown_args):
             raise CommandError("No valid .zap file provided")
 
         if args.output:
-            output_path = args.output
+            output_path = zap_file_path.parent / args.output
```
Damian-Nordic, line 52: Is it actually intuitive that it is relative to the ZAP file? Most tools assume that relative path is relative to the current directory, so it's worth at least documenting that in the help text. Also, in such a case you should probably make sure the provided path isn't absolute.

maciejbaczmanski: You're right @Damian-Nordic . Just found that we were simply missing `.absolute()` call, and that's why passing a relative path was causing it to be searched by zap generator. When we pass an absolute path, `west zap-generate` can be called either with a path relative to current dir or an absolute path.

Damian-Nordic (approved):

scripts/west/zap_generate.py
Damian-Nordic: Thanks for checking!

kkasperczyk-no (approved):

PR #464, opened 2024-07-22
[nrf toup] [nrfconnect] Implement IPv6 agnostic L2 network state getters

markaj-nordic:
* Added new API for generic IPv6 connectivity checks
* This patch allows the application layer to limit the amount of pre-processor logic (CONFIG_NET_L2_OPENTHREAD/CONFIG_CHIP_WIFI) needed to configure the

### DIFF
--- a/src/platform/nrfconnect/ConnectivityManagerImpl.h
+++ b/src/platform/nrfconnect/ConnectivityManagerImpl.h
@@ -81,6 +81,32 @@ class ConnectivityManagerImpl final : public ConnectivityManager,
     // the implementation methods provided by this class.
     friend class ConnectivityManager;
 
+public:
+    // Generic network status checkers
+    bool IsIPv6NetworkEnabled()
+    {
+        return false
+#if CHIP_DEVICE_CONFIG_ENABLE_THREAD
+            || IsThreadEnabled()
+#endif
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFI
+            || IsWiFiStationEnabled()
+#endif
+            ;
+    };
+
+    bool IsIPv6NetworkProvisioned()
+    {
+        return false
+#if CHIP_DEVICE_CONFIG_ENABLE_THREAD
+            || IsThreadProvisioned()
+#endif
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFI
+            || IsWiFiStationProvisioned()
+#endif
+            ;
+    }
+
 private:
     // ===== Members that implement the ConnectivityManager abstract interface.
 


src/platform/nrfconnect/ConnectivityManagerImpl.h
```diff
@@ -17,6 +17,10 @@
 
 #pragma once
 
+#if !CHIP_DEVICE_CONFIG_ENABLE_THREAD && !CHIP_DEVICE_CONFIG_ENABLE_WIFI
+#error "At least one IPv6 networking backend must be enabled!"
```
Damian-Nordic, line 21: I think this will cause native_posix test failures. I suggest avoiding this assumption.

src/platform/nrfconnect/ConnectivityManagerImpl.h
```diff
@@ -81,6 +85,44 @@ class ConnectivityManagerImpl final : public ConnectivityManager,
...
 
+public:
+    // Generic network status checkers
+    bool IsIPv6NetworkEnabled()
+    {
+#if CHIP_DEVICE_CONFIG_ENABLE_THREAD
+        return IsThreadEnabled();
+#elif CHIP_DEVICE_CONFIG_ENABLE_WIFI
+        return IsWiFiStationEnabled();
+#endif
```
Damian-Nordic, line 96: For thread + Wi-Fi configuration, these checks should be ORed, right?

```suggestion
        return false
#if CHIP_DEVICE_CONFIG_ENABLE_THREAD
            || IsThreadEnabled();
#endif
#if CHIP_DEVICE_CONFIG_ENABLE_WIFI
            || IsWiFiStationEnabled();
#endif
            ;
```

src/platform/nrfconnect/ConnectivityManagerImpl.h
```diff
@@ -81,6 +85,44 @@ class ConnectivityManagerImpl final : public ConnectivityManager,
...
+    };
+
+    bool IsIPv6NetworkApplicationControlled()
```
Damian-Nordic, line 99: Is this used anywhere?

markaj-nordic: Not really, I added it for completeness sake. But right, it's not needed for now.

src/platform/nrfconnect/ConnectivityManagerImpl.h
markaj-nordic: Right, will remove it.

src/platform/nrfconnect/ConnectivityManagerImpl.h
markaj-nordic: The same for `IsIPv6NetworkAttached()`

src/platform/nrfconnect/ConnectivityManagerImpl.h
markaj-nordic: Ok, that's right. But this applies to the `IsEnable()` only, right? When it comes to `IsProvisioned()` we cannot have both Thread and Wi-Fi provisioned at the same time.

Damian-Nordic: Actually, we can, but even if we couldn't, you would need to support the case where we have both Thread + Wi-Fi support **enabled**, so it's still worth ORing the return values in such a case.

markaj-nordic: Yes, definitely we should do OR in the case of IsEnable(). The only thing I have doubts about is the possibility of simultaneous provisioning of both backends. Is it really a case for our samples? It sounds more like a TBR case to me.

Damian-Nordic: For our samples, no, but it is possible to have two NetworkCommissioning instances for Thread + WiFi and have both provisioned. What would you use instead of OR even if this wasn't possible?

markaj-nordic: Discussed f2f.

src/platform/nrfconnect/ConnectivityManagerImpl.h
```diff
@@ -81,6 +81,32 @@ class ConnectivityManagerImpl final : public ConnectivityManager,
...
+public:
+    // Generic network status checkers
+    bool IsIPv6NetworkEnabled()
+    {
+        return false
+#if CHIP_DEVICE_CONFIG_ENABLE_THREAD
+            || IsThreadEnabled();
+#endif
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFI
+        || IsWiFiStationEnabled();
```
Damian-Nordic, line 93:
```suggestion
        || IsWiFiStationEnabled()
```

kkasperczyk-no (approved):

ArekBalysNordic (approved):

PR #467, opened 2024-07-31
[nrfconnect] Introduced Kconfig to configure CHIP stack size

kkasperczyk-no:
There is not Kconfig that would allow to set the CHIP task size, what is not convenient, as other stacks can be configured using Kconfig.

manfiest-PR-skip


Damian-Nordic (approved):
Add `[nrf topu]` to the commit message.

markaj-nordic (approved):

PR #469, opened 2024-08-05
[nrf noup] wifi: Migrate to nRF70 upstream

krish2718:
nRF70 support moved upstream, only Kconfigs have been renamed from nrf700x->nrf70 and WPA_SUPP->WIFI_NM_WPA_SUPPLICANT


config/nxp/chip-module/Kconfig.features
```diff
@@ -24,11 +24,11 @@ config CHIP_WIFI
 	default y
 	select WIFI
 	select WIFI_NXP
-	select WPA_SUPP
```
markaj-nordic, line 27: I suppose this change is not the result of adding the driver to the Zephyr but rather the refactoring needed to align everything? By that I mean that the WPA_SUPP_AP was defined in the hostap not in the driver (and the hostap have been integrated in Zephyr for a while already). Anyway, I wonder if we should first upstream the patch that affects platforms other than nrfconnect. @LuDuda, @kkasperczyk-no any thoughts here?

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -252,35 +252,35 @@ endif # NET_L2_OPENTHREAD
 
 if CHIP_WIFI
 
-choice WPA_SUPP_LOG_LEVEL_CHOICE
-    default WPA_SUPP_LOG_LEVEL_ERR
+choice WIFI_NM_WPA_SUPPLICANT_LOG_LEVEL_CHOICE
+    default WIFI_NM_WPA_SUPPLICANT_LOG_LEVEL_ERR
```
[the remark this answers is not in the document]
krish2718: Network manager https://docs.zephyrproject.org/latest/doxygen/html/group__wifi__nm.html

config/nxp/chip-module/Kconfig.features
krish2718: nR70 now uses hostap from the upstream where the renaming was done.

krish2718:
please review

config/nxp/chip-module/Kconfig.features
markaj-nordic: Yes, my only concern was if the NXP already used the hostap from Zephyr upstream that includes this change.

markaj-nordic (approved):

bjarki-andreasen (approved):

krish2718:
These changes are merged as part of upmerge, so, closing.

PR #470, opened 2024-08-30
[nrf toup][nrfconnect] Added BT reassembly config to BT SMP samples

HubertHugo:
This PR adds BT reassembly config to SMP DFU samples as it is recommended

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig.features
+++ b/config/nrfconnect/chip-module/Kconfig.features
@@ -158,6 +158,7 @@ config CHIP_DFU_OVER_BT_SMP
 	# Enable custom SMP request to erase settings partition.
 	select MCUMGR_GRP_ZBASIC
 	select MCUMGR_GRP_ZBASIC_STORAGE_ERASE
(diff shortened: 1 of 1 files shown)


markaj-nordic:
Maybe it's better to first add it to upstream and then pull in as [nrf fromtree.]

HubertHugo:
> Maybe it's better to first add it to upstream and then pull in as [nrf fromtree.]

@markaj-nordic I don't know what are pluses and minuses of this approach so I can change it if you guys agree on that.

markaj-nordic:
> > Maybe it's better to first add it to upstream and then pull in as [nrf fromtree.]
> 
> @markaj-nordic I don't know what are pluses and minuses of this approach so I can change it if you guys agree on that.

For me the rule of thumb for [toup] is the need of urgent patch/fix in our SDK. This change is just a minor improvement, so I'd suggest to go upmerge first. On the other hand, it's just our platform being affected, so I maybe I'm too picky.

PR #473, opened 2024-09-05
[nrf fromtree][nrfconnect] Added BT reassembly config as default for …

HubertHugo:
This PR adds BT reassembly config for all SMP DFU samples as it is recommended

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig.features
+++ b/config/nrfconnect/chip-module/Kconfig.features
@@ -158,6 +158,7 @@ config CHIP_DFU_OVER_BT_SMP
 	# Enable custom SMP request to erase settings partition.
 	select MCUMGR_GRP_ZBASIC
 	select MCUMGR_GRP_ZBASIC_STORAGE_ERASE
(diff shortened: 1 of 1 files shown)


markaj-nordic (approved):

maciejbaczmanski (approved):

ArekBalysNordic (approved):

config/nrfconnect/chip-module/Kconfig.features
```diff
@@ -158,6 +158,7 @@ config CHIP_DFU_OVER_BT_SMP
 	# Enable custom SMP request to erase settings partition.
 	select MCUMGR_GRP_ZBASIC
 	select MCUMGR_GRP_ZBASIC_STORAGE_ERASE
+	select MCUMGR_TRANSPORT_BT_REASSEMBLY
```
kkasperczyk-no, line 161: Could we change this from select to imply? I can see there might be some advantages from using it, but I would not say this is crucial for SMP to work and we have to force it without allowance to disable it.

PR #475, opened 2024-09-11
[nrf toup][nrfconnect] Refactor ExternalFlashManager

maciejbaczmanski:
Make `ExternalFlashManager` a singleton and move `DoAction` implementation to `ExternalFlashManager.cpp`

### DIFF
--- a/src/platform/nrfconnect/ExternalFlashManager.h
+++ b/src/platform/nrfconnect/ExternalFlashManager.h
@@ -31,9 +31,19 @@ public:
         SLEEP
     };
 
-    virtual ~ExternalFlashManager() {}
+    // Not copyable or movable
+    ExternalFlashManager(ExternalFlashManager &&)                  = delete;
+    ExternalFlashManager & operator=(const ExternalFlashManager &) = delete;
+    ExternalFlashManager(const ExternalFlashManager &)             = delete;
+    ExternalFlashManager & operator=(ExternalFlashManager &&)      = delete;
 
-    virtual void DoAction(Action aAction)
+    static ExternalFlashManager& GetInstance()
+    {
+        static ExternalFlashManager sExternalFlashManager;
+        return sExternalFlashManager;
+    }
+
+    void DoAction(Action aAction)
     {
 #if defined(CONFIG_PM_DEVICE) && defined(CONFIG_NORDIC_QSPI_NOR)
         // utilize the QSPI driver sleep power mode
@@ -45,6 +55,10 @@ public:
         }
 #endif
     }
+
+private:
+    // Singleton Object
+    ExternalFlashManager() = default;
 };
 
 } // namespace DeviceLayer


src/platform/nrfconnect/ExternalFlashManager.h
```diff
@@ -31,20 +30,20 @@ class ExternalFlashManager
...
-        // utilize the QSPI driver sleep power mode
-        const auto * qspi_dev = DEVICE_DT_GET(DT_INST(0, nordic_qspi_nor));
-        if (device_is_ready(qspi_dev))
-        {
-            const auto requestedAction = Action::WAKE_UP == aAction ? PM_DEVICE_ACTION_RESUME : PM_DEVICE_ACTION_SUSPEND;
-            (void) pm_device_action_run(qspi_dev, requestedAction); // not much can be done in case of a failure
-        }
-#endif
-    }
+    static ExternalFlashManager& getInstance() { return sExternalFlashManager; }
```
kkasperczyk-no, line 39: Usually we use UpperCamelCase for methods.
```suggestion
    static ExternalFlashManager& GetInstance() { return sExternalFlashManager; }
```

kkasperczyk-no: Also alternatively, you could get rid of sExternalFlashManager field this way (but this is a minor and depends on the taste :) ) 
```suggestion
    static ExternalFlashManager& getInstance() { 
      static ExternalFlashManager sExternalFlashManager;
      return sExternalFlashManager; 
      }
```

src/platform/nrfconnect/ExternalFlashManager.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <zephyr/device.h>
+#include <zephyr/pm/device.h>
+
+namespace chip {
+namespace DeviceLayer {
+
+ExternalFlashManager ExternalFlashManager::sExternalFlashManager;
+
+void ExternalFlashManager::DoAction(Action aAction)
```
kkasperczyk-no, line 27: Just to my curiosity, why do we have to move it to .cpp file?

kkasperczyk-no (approved):
I left few minor comments, but lgtm overall.

src/platform/nrfconnect/ExternalFlashManager.h
maciejbaczmanski: Right, with these suggestions, `.cpp` file is not needed anymore, so I've deleted it as well :) thanks!

src/platform/nrfconnect/ExternalFlashManager.h
```diff
@@ -31,9 +31,19 @@ class ExternalFlashManager
...
     };
 
-    virtual ~ExternalFlashManager() {}
+    // Not copyable or movable
+    ExternalFlashManager(ExternalFlashManager &&)                  = delete;
+    ExternalFlashManager & operator=(const ExternalFlashManager &) = delete;
+    ExternalFlashManager(const ExternalFlashManager &)             = delete;
+    ExternalFlashManager & operator=(ExternalFlashManager &&)      = delete;
 
-    virtual void DoAction(Action aAction)
```
markaj-nordic, line 36: Since we remove the possibility of overriding this functionality, maybe it's better to move ExternalFlashManager to the sdk-nrf scope as it assumes the external flash is using QSPI bus which is specific to our DKs. That might not always be the case in end products, so we could at least make it more transparent and application- scoped.

markaj-nordic, line 36: Since we remove the possibility of overriding this functionality, maybe it's better to move ExternalFlashManager to the sdk-nrf scope as it assumes the external flash is using QSPI bus which is specific to our DKs. That might not always be the case in end products, so we could at least make it more transparent and application- scoped.

src/platform/nrfconnect/ExternalFlashManager.h
markaj-nordic: Note that Mayer's singleton is elegant and thread safe but it usually comes with additional overhead due to locking, which might be good as a matter of fact. So, in essence these two are not exactly the same thing and it's not only a matter of a taste :)

src/platform/nrfconnect/ExternalFlashManager.h
maciejbaczmanski: Discussed offline- we've decided to keep the implementation as is for now as ExternalFlashManager is used in core as well. Created a follow-up task to investigate how we can make this modular and allow other implementations (like SPI): KRKNWK-19452

PR #480, opened 2024-09-16
[nrf noup] Mark `CHIP_ICD_LIT_SUPPORT` experimental

maciejbaczmanski:
LIT functionality is provisional for Matter 1.3.
Set it as experimental for now.


kkasperczyk-no (commented):
I think we should add experimental to UAT and CIP as well, because they can be optionally enabled for SIT.

maciejbaczmanski:
> I think we should add experimental to UAT and CIP as well, because they can be optionally enabled for SIT.

right. Added

doublemis1 (approved):

kkasperczyk-no (approved):

LuDuda (approved):

PR #487, opened 2024-10-08
[nrf noup] Shedule factory reset in Server using Matter Shell command

ArekBalysNordic:
It is better to call the ScheduleFactoryReset method from the Server because it also removes all fabrics and emits the Leave event whereas calling InitiateFactoryReset() removes only the persistent storage entries.

In this way, the Matter Shell command works in the same way as other factory reset sources.

We cannot upmerge it in an easy way because many vendors do not use chip server and there are multiple building issues.

### DIFF
--- a/src/lib/shell/commands/Device.cpp
+++ b/src/lib/shell/commands/Device.cpp
@@ -15,6 +15,7 @@
  *    limitations under the License.
  */
 
+#include <app/server/Server.h>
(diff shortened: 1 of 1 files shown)


LuDuda (approved):
Thanks :+1:

maciejbaczmanski (approved):

PR #488, opened 2024-10-08
Fix wifi heap

krish2718:
* WPA supplicant now uses k_heap
* nRF70 driver now allocates higher heap, so, suppress the warning and extra allocation


config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -54,10 +54,12 @@ config SYSTEM_WORKQUEUE_STACK_SIZE
...
 
 config HEAP_MEM_POOL_SIZE
-    default 80000 if CHIP_WIFI
+    default 110720 if CHIP_WIFI
+
+config HEAP_MEM_POOL_IGNORE_MIN
+    default y if CHIP_WIFI
 
 config CHIP_MALLOC_SYS_HEAP_SIZE
-    default 30720 if CHIP_WIFI
```
Damian-Nordic, line 60: You seem to assume that only Wi-Fi has used this heap, which is not true. `config CHIP_MALLOC_SYS_HEAP_SIZE` defaults to 12kB, so by adding 30720 to `HEAP_MEM_POOL_SIZE` and removing this default, you effectively increase the RAM usage by 12kB.

Do you have an idea how much the supplicant actually needs?

krish2718: > You seem to assume that only Wi-Fi has used this heap, which is not true. 

Yes, I assume CHIP_WIFI is only for Wi-Fi (wpa supplicant). WPA supplicant needs 30000 bytes of heap,

Damian-Nordic: I meant that `CHIP_MALLOC_SYS_HEAP_SIZE` defines the size of the common libc heap and part of that is used for Matter itself, so if it was previously set to `30720B` it means that not all was taken by the WPA supplicant in our case.

krish2718: Ah understood, but as WPA supplicant needs a minimum of 30000, isn't the extra heap 720? How is it 12Kb?

LuDuda: In our stress tests for Thread it seemed we achieved around 5k of usage, and set it to 8k to have some buffer. For Wi-Fi it was previously around 27k for both Wi-Fi and Matter. But surely we might not reproduced the worst case scenario for Wi-Fi. If 30k is recommendation ofc let's use this number (but indeed it increases RAM usage by 12k if i'm correct - though for NCS 2.7.0 we had more than 150k free RAM - so additional 8 may fly for nRF5340 if it's needed).

So I would keep `CHIP_MALLOC_SYS_HEAP_SIZE` as 8k for Thread and 12k for Wi-Fi. Now since Wi-Fi uses only one heap, we might be able to profile this config for Wi-Fi easier (potentially 8k would be enough as well).

@Damian-Nordic what do you think?

Damian-Nordic: @krish2718 The original `CHIP_MALLOC_SYS_HEAP_SIZE` config (defined in a different file) defaults to 12kB so by removing the default from here, you effectively fall back to 12kB.
But ok, if @LuDuda says we're not short on memory for Matter over WiFi and are OK with the 12kB increase, let's leave it as it is :)

Damian-Nordic (approved):

PR #491, opened 2024-10-09
[nrf noup] Increase stack sizes for devices that runs CRACEN

ArekBalysNordic:
CRACEN requires more stack memory for all threads that use many cryptographic operations.


LuDuda (approved):

ArekBalysNordic:
Seems the issue we faced recently is caused by something different than a lack of stack space.

PR #492, opened 2024-10-10
[nrf toup] examples: align nRF5340 IPC

gmarull:
nRF5340 IPC now uses icbmsg, so align all overlays accordingly.


rlubos (approved):

ArekBalysNordic (commented):
None of these samples are tested in NCS, so all changes here are redundant. If this PR is merged, we will remove it anyway during the next bumping the NCS version in the Matter's upstream repo.

PR #493, opened 2024-10-11
Add ZMS to Zephyr port 

ArekBalysNordic:
Add ZMS to Zephyr port, Enable ZMS in nrfconnect if RRAM is in use.


config/nrfconnect/chip-module/Kconfig
```diff
@@ -325,8 +329,16 @@ endif # CHIP_FACTORY_DATA_BUILD
...
+	depends on NVS
 	depends on !CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
 
+config CHIP_FACTORY_RESET_ERASE_ZMS
+	bool
+	default y
+	depends on ZMS
+	depends on !CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
+
+
```
maciejbaczmanski, line 341:
```suggestion
```

config/zephyr/Kconfig
```diff
@@ -479,6 +479,18 @@ config CHIP_FACTORY_RESET_ERASE_NVS
 	  configuration is supposed to be cleared on a factory reset, including
 	  device-specific entries.
 
+config CHIP_FACTORY_RESET_ERASE_ZMS
+	bool "Erase NVS flash pages on factory reset"
```
maciejbaczmanski, line 483: Should prompt and help be exactly the same as for `CHIP_FACTORY_RESET_ERASE_NVS` ?

ArekBalysNordic: right, it shouldn't, thanks!

config/zephyr/Kconfig
```diff
@@ -477,7 +477,19 @@ config CHIP_FACTORY_RESET_ERASE_NVS
 	  regain the original storage performance if any firmware issue has brought
 	  it to an unexpected state. For this reason, set this option if the entire
 	  configuration is supposed to be cleared on a factory reset, including
-	  device-specific entries.
+	  device-specific entries. It makes sense only for NVS filesystem.
+
+config CHIP_FACTORY_RESET_ERASE_ZMS
```
LuDuda, line 482: It's fine for now and will accept this, however, what do you think to have one Matter config like `CHIP_FACTORY_RESET_PARTITION_ERASE` which would use ZMS or NVS underneath based on the selected backend?

config/zephyr/Kconfig
```diff
@@ -477,7 +477,19 @@ config CHIP_FACTORY_RESET_ERASE_NVS
...
+	bool "Erase ZMS flash pages on factory reset"
```
LuDuda, line 483:
```suggestion
	bool "Erase ZMS non-volatile pages on factory reset"
```

config/zephyr/Kconfig
```diff
@@ -477,7 +477,19 @@ config CHIP_FACTORY_RESET_ERASE_NVS
...
+	depends on SETTINGS_ZMS
```
LuDuda, line 484: Do we have protection on setting this config when DAC migration is turned on? I mean we should not allow for setting this config if DAC migration is enabled.

ArekBalysNordic: Yes I've added it to the nrfconnect's .Kconfig file, but please double check it :)

config/zephyr/Kconfig
ArekBalysNordic: Yeah but it impacts many platforms that use Zephyr... We can do it, but not so easily. :/ I think we can create a follow up task for that after the release. :)

kkasperczyk-no: Yeah, I also think that using two different Kconfigs for that is not the best idea. Why is it not that easy? Isn't it just about changing CHIP_FACTORY_RESET_ERASE_NVS->CHIP_FACTORY_RESET_ERASE_SETTINGS and then adding depends on depends on SETTINGS_NVS || SETTINGS_ZMS? Then we can use zms_clear or nvs_clear in the cpp file based on SETTINGS_NVS/SETTINGS_ZMS selection, because it's a choice. I don't think it would affect anyone much unless he will use SETTINGS_ZMS.

config/nrfconnect/chip-module/Kconfig
```diff
@@ -322,9 +326,9 @@ config CHIP_DEVICE_GENERATE_ROTATING_DEVICE_UID
 endif # CHIP_FACTORY_DATA_BUILD
 
 # See config/zephyr/Kconfig for full definition
-config CHIP_FACTORY_RESET_ERASE_NVS
-	bool
+config CHIP_FACTORY_RESET_ERASE_SETTINGS
 	default y
+	depends on NVS || KVS
```
kkasperczyk-no, line 331: KVS?

config/zephyr/Kconfig
```diff
@@ -44,7 +44,7 @@ menuconfig CHIP
 	imply HWINFO
 	imply FLASH
 	imply FLASH_MAP
-	imply NVS
+	imply NVS if !CHIP_NRF_PLATFORM
```
kkasperczyk-no, line 47: Why not:

	imply NVS if SETTINGS_NVS
	imply ZMS if SETTINGS_ZMS

?

src/platform/Zephyr/ConfigurationManagerImpl.cpp
```diff
@@ -194,20 +198,24 @@ void ConfigurationManagerImpl::DoFactoryReset(intptr_t arg)
...
+        status = zms_clear(static_cast<zms_fs *>(storage));
+#endif
     }
-
     if (status)
     {
         ChipLogError(DeviceLayer, "Factory reset failed: %d", status);
     }
-#else
+#else  // CONFIG_SETTINGS_NVS || CONFIG_SETTINGS_ZMS
```
kkasperczyk-no, line 217: Redundant comment.

config/zephyr/Kconfig
kkasperczyk-no: Discussed f2f.

src/platform/nrfconnect/FactoryDataProvider.cpp
```diff
@@ -160,6 +160,10 @@ CHIP_ERROR FactoryDataProvider<FlashFactoryData>::MoveDACPrivateKeyToSecureStora
         }
 
 #ifdef CONFIG_CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
+#ifdef CONFIG_CHIP_FACTORY_RESET_ERASE_SETTINGS
+#error "Do not use both CONFIG_CHIP_FACTORY_RESET_ERASE_SETTINGS and CONFIG_CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY kconfig options " \
+       "because you will lose the DAC private key permanently from the device."
```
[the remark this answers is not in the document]
wiba-nordic: "lose", not "loose".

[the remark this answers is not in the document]
kkasperczyk-no: Ah right :)

doublemis1:
Can we somehow highlight what Configs have been changed?
Clients may use them in their configuration and during migration bump into issues with config renaming.
Maybe add it to migration guideline

PR #499, opened 2024-10-23
[nrf toup] Factory data partition location change

kg-nordicsemi:
- factory data can be placed before or after settings partition
-added static_asserts, which checks whether factory data overlaps with settings partition
-added Kconfig which allow to choose the location of settings and factory data partition

### DIFF
--- a/src/platform/nrfconnect/FactoryDataProvider.h
+++ b/src/platform/nrfconnect/FactoryDataProvider.h
@@ -69,7 +69,7 @@ struct InternalFlashFactoryData
     // 2) it does not exceed the settings partition start address
     // Note that this block can overlap with app partition but this is not a problem since we do not aim to modify
     // the application code at runtime anyway.
-    constexpr size_t FactoryDataBlockBegin()
+    static constexpr size_t FactoryDataBlockBegin()
     {
         // calculate the nearest multiple of CONFIG_FPROTECT_BLOCK_SIZE smaller than FACTORY_DATA_ADDRESS
         return FACTORY_DATA_ADDRESS & (-CONFIG_FPROTECT_BLOCK_SIZE);
@@ -81,11 +81,19 @@ struct InternalFlashFactoryData
         // and make sure we do not overlap with settings partition
         constexpr size_t kFactoryDataBlockEnd =
             (FACTORY_DATA_ADDRESS + FACTORY_DATA_SIZE + CONFIG_FPROTECT_BLOCK_SIZE - 1) & (-CONFIG_FPROTECT_BLOCK_SIZE);
-        static_assert(kFactoryDataBlockEnd <= PM_SETTINGS_STORAGE_ADDRESS,
+
+        // Only the partition that is protected by fprotect must be aligned to fprotect block size
+        constexpr size_t kSettingsBlockEnd = PM_SETTINGS_STORAGE_ADDRESS + PM_SETTINGS_STORAGE_SIZE;
+
+        constexpr bool kOverlapsCheck =
+            (kSettingsBlockEnd <= FactoryDataBlockBegin()) || (kFactoryDataBlockEnd <= PM_SETTINGS_STORAGE_ADDRESS);
+
+        static_assert(kOverlapsCheck,
                       "FPROTECT memory block, which contains factory data"
                       "partition overlaps with the settings partition."
                       "Probably your settings partition size is not a multiple"
                       "of the atomic FPROTECT block size of " TO_STR(CONFIG_FPROTECT_BLOCK_SIZE) "kB");
+
         return kFactoryDataBlockEnd - FactoryDataBlockBegin();
     }
 #undef TO_STR
@@ -100,7 +108,10 @@ struct InternalFlashFactoryData
 #endif // if CONFIG_FPROTECT
     }
 #else
-    CHIP_ERROR ProtectFactoryDataPartitionAgainstWrite() { return CHIP_ERROR_NOT_IMPLEMENTED; }
+    CHIP_ERROR ProtectFactoryDataPartitionAgainstWrite()
+    {
+        return CHIP_ERROR_NOT_IMPLEMENTED;
+    }
 #endif
 };
 


ArekBalysNordic (commented):
Be aware of this note: https://github.com/nrfconnect/sdk-connectedhomeip/pull/499/files#diff-4521c3129b9a1ff0496c09d5af49e450c8e50b97a538845bcbae1d386f060536R64

src/platform/nrfconnect/FactoryDataProvider.h
```diff
@@ -72,15 +72,26 @@ struct InternalFlashFactoryData
...
     constexpr size_t FactoryDataBlockSize()
     {
-        // calculate the factory data end address rounded up to the nearest multiple of CONFIG_FPROTECT_BLOCK_SIZE
+        // calculate the factory data and settings end address rounded up to the nearest multiple of CONFIG_FPROTECT_BLOCK_SIZE
         // and make sure we do not overlap with settings partition
         constexpr size_t kFactoryDataBlockEnd =
             (FACTORY_DATA_ADDRESS + FACTORY_DATA_SIZE + CONFIG_FPROTECT_BLOCK_SIZE - 1) & (-CONFIG_FPROTECT_BLOCK_SIZE);
-        static_assert(kFactoryDataBlockEnd <= PM_SETTINGS_STORAGE_ADDRESS,
+
+        constexpr size_t SettingsBlockEnd =
```
kkasperczyk-no, line 80: @ArekBalysNordic please correct me if I'm wrong, but I think that the settings partition does not need to be aligned to the fprotect block size, so the end address is just start address + size, right?

src/platform/nrfconnect/FactoryDataProvider.h
```diff
@@ -72,15 +72,26 @@ struct InternalFlashFactoryData
...
+            (PM_SETTINGS_STORAGE_ADDRESS + PM_SETTINGS_STORAGE_SIZE + CONFIG_FPROTECT_BLOCK_SIZE - 1) &
+            (-CONFIG_FPROTECT_BLOCK_SIZE);
+
+        constexpr size_t kFactoryDataBlockBegin = FACTORY_DATA_ADDRESS & (-CONFIG_FPROTECT_BLOCK_SIZE);
+
+        constexpr bool overlapsCheck =
+            (SettingsBlockEnd <= kFactoryDataBlockBegin) || (kFactoryDataBlockEnd <= PM_SETTINGS_STORAGE_ADDRESS);
```
kkasperczyk-no, line 87: I'm not sure that this is correct. The problem that we consider here is the fact that fprotect has some fixed block size and we want to use only on factory data (not on settings). There are two issues possible:

1. The factory data is before the settings and its size is not aligned to the fprotect block size, so bigger range than expected will be covered by fprotect. In such case we have to check if factory data start+ factory data size + alignment to fprotect block size will not overlap with the start address of settings.
2. The settings are before the factory data and the factory data size is not aligned. If the factory data is at the end of memory, the fprotect will do the alignment using the space before the factory data (because after that there is no memory left). In such case we have to check if memory end - (factory data size + alignment) does not overlap with the settings end.

src/platform/nrfconnect/FactoryDataProvider.h
ArekBalysNordic: That's right. Only the partition that is protected by fprotect must be aligned to fprotect block size. So we need to take care of factory data only. We need to be aware to not overlap settings partition by fprotect. Keep in mind that the project block size varies depending on our platform.

src/platform/nrfconnect/FactoryDataProvider.h
```diff
@@ -76,11 +76,21 @@ struct InternalFlashFactoryData
         // and make sure we do not overlap with settings partition
         constexpr size_t kFactoryDataBlockEnd =
             (FACTORY_DATA_ADDRESS + FACTORY_DATA_SIZE + CONFIG_FPROTECT_BLOCK_SIZE - 1) & (-CONFIG_FPROTECT_BLOCK_SIZE);
-        static_assert(kFactoryDataBlockEnd <= PM_SETTINGS_STORAGE_ADDRESS,
+
+        constexpr size_t kFactoryDataBlockBegin = FACTORY_DATA_ADDRESS & (-CONFIG_FPROTECT_BLOCK_SIZE);
```
kkasperczyk-no, line 80: @ArekBalysNordic could you verify the alignment of the factory data beginning? TBH I don't know where this magic with & and - fprotect block size comes from.

kkasperczyk-no (approved):
LGTM overall, please consult with @ArekBalysNordic  that alignment is correct.

src/platform/nrfconnect/FactoryDataProvider.h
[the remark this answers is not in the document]
kg-nordicsemi: Because if I do this like this: constexpr size_t kFactoryDataBlockBegin = FactoryDataBlockBegin();                                         There is an error:                                                                                                                                                                                              
                                                                                                                                                                                              In member function 'constexpr size_t chip::DeviceLayer::InternalFlashFactoryData::FactoryDataBlockSize()':
../../../../../../../../modules/lib/matter/src/platform/nrfconnect/FactoryDataProvider.h:81:73: error: 'this' is not a constant expression
   81 |         constexpr size_t kFactoryDataBlockBegin = FactoryDataBlockBegin();
      |                                                                         ^
../../../../../../../../modules/lib/matter/src/platform/nrfconnect/FactoryDataProvider.h:89:23: error: non-constant condition for static assertion
   89 |         static_assert(kOverlapsCheck,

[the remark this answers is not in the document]
kg-nordicsemi: Because if I do this like this: constexpr size_t kFactoryDataBlockBegin = FactoryDataBlockBegin();
there is an following error:
image

ArekBalysNordic (approved):
Please rebase the related sdk-nrf manifest PR to see whether it works before merging.

PR #504, opened 2024-10-30
[nrf noup] Provide a workaround for Factory Data support on nRF54H

ArekBalysNordic:
Until we do not have a proper solution for merging the factory data hex file with the firmware hex file, we need to flash the factory data manually. This commit allows doing that on nRF54H20 DK.

### DIFF
--- a/config/nrfconnect/chip-module/generate_factory_data.cmake
+++ b/config/nrfconnect/chip-module/generate_factory_data.cmake
@@ -141,6 +141,17 @@ else()
     endif()
     dt_reg_addr(factory_data_addr PATH ${factory_data_alias})
     dt_reg_size(factory_data_size PATH ${factory_data_alias})
+
+    # If the location of the factory data is defined as an alias,
+    # add the base address to the factory data offset
+    # to get the proper offset calculated from the base memory location.
+    dt_alias(factory_data_location PROPERTY "factory-data-location")
+    if(factory_data_location)
+        dt_node_exists(factory_data_location_exists PATH "${factory_data_location}")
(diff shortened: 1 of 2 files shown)


config/nrfconnect/chip-module/generate_factory_data.cmake
```diff
@@ -141,6 +141,17 @@ else()
     endif()
     dt_reg_addr(factory_data_addr PATH ${factory_data_alias})
     dt_reg_size(factory_data_size PATH ${factory_data_alias})
+
+    # If the location of the factory data is defined as an alias,
+    # add the base address to the factory data offset
+    # to get the proper offset calcualted from the base memory location.
```
kkasperczyk-no, line 147:
```suggestion
    # to get the proper offset calculated from the base memory location.
```

src/platform/nrfconnect/FactoryDataProvider.h
```diff
@@ -36,7 +36,12 @@
 #else
 #include <zephyr/storage/flash_map.h>
 #define FACTORY_DATA_SIZE DT_REG_SIZE(DT_ALIAS(factory_data))
+#define FACTORY_DATA_LOCATION_ADDRESS DT_REG_ADDR(DT_ALIAS(factory_data_location))
```
kkasperczyk-no, line 39: Is it going to pass the compilation if `factory_data_location` does not exist?

kkasperczyk-no: Hmm, I've done some tests locally and it seems to work.

kkasperczyk-no (approved):

PR #507, opened 2024-11-05
[nrf toup] Reduce Packet Buffer Pool size to 11

adigie:


src/platform/nrfconnect/SystemPlatformConfig.h
```diff
@@ -48,7 +48,7 @@ struct ChipDeviceEvent;
 #define CHIP_SYSTEM_CONFIG_USE_SOCKETS 1
 
 #ifndef CONFIG_ARCH_POSIX
-#define CHIP_SYSTEM_CONFIG_PACKETBUFFER_POOL_SIZE 15
+#define CHIP_SYSTEM_CONFIG_PACKETBUFFER_POOL_SIZE 11
```
Damian-Nordic, line 51: Is this selected based on the top usage in tests?

Damian-Nordic (approved):

src/platform/nrfconnect/SystemPlatformConfig.h
Damian-Nordic: Also, please fill in this info in the commit description :)

LuDuda: As this is not yet targeted for 2.8.0 and we have a bit of time - what do you think to introduce a KConfig for this configuration?

adigie: > As this is not yet targeted for 2.8.0 and we have a bit of time - what do you think to introduce a KConfig for this configuration?

Good idea. 


> Is this selected based on the top usage in tests?

Yes, highest observed pool usage was 10. I updated commit description.

config/nrfconnect/chip-module/Kconfig
```diff
@@ -459,4 +459,10 @@ config CHIP_ENABLE_BDX_LOG_TRANSFER
 	help
 	  Enables the BDX protocol for diagnostics log transfer purposes.
 
+config CHIP_SYSTEM_PACKETBUFFER_POOL_SIZE
+  int "Packet buffer pool size"
+  default 11
+  help
+    Total number of packet buffers.
```
ArekBalysNordic, line 466: I wonder if we can describe it a little bit more. For example, describe the purposes for which those packet buffers will be used.

ArekBalysNordic (approved):
Looks good. Please squash commits before merging.

PR #508, opened 2024-11-05
[nrf toup] [nrfconnect] disable `FPROTECT` for nRF54L15

maciejbaczmanski:
`FPROTECT` should be disabled for nRF54L15 in app to allow correctly protecting whole region of mcuboot

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig
+++ b/config/nrfconnect/chip-module/Kconfig
@@ -228,7 +228,7 @@ config CHIP_FACTORY_DATA_ROTATING_DEVICE_UID_MAX_LEN
 config CHIP_FACTORY_DATA_WRITE_PROTECT
 	bool "Enable Factory Data write protection"
 	select FPROTECT
-	depends on CHIP_FACTORY_DATA
+	depends on CHIP_FACTORY_DATA && !SOC_SERIES_NRF54LX
 	default y
 	help
 		Enables the write protection of the Factory Data partition in the flash memory.


adigie (approved):

config/nrfconnect/chip-module/Kconfig
```diff
@@ -228,7 +228,7 @@ config CHIP_FACTORY_DATA_ROTATING_DEVICE_UID_MAX_LEN
 config CHIP_FACTORY_DATA_WRITE_PROTECT
 	bool "Enable Factory Data write protection"
 	select FPROTECT
-	depends on CHIP_FACTORY_DATA
+	depends on CHIP_FACTORY_DATA && !SOC_SERIES_NRF54LX
```
kkasperczyk-no, line 231: One small comment to be considered. Note that before the FPROTECT was set using `select`, so we were sure it is enabled. After this change it is enabled as a default in mcuboot, so someone may disable it and we will end up having this Kconfig enabled, but without having fprotect enabled.

maciejbaczmanski: but it was set for application. now we do not want `FPROTECT` nor `CHIP_FACTORY_DATA_WRITE_PROTECT` to be set in application for NRF54L15, hence dependency on `CHIP_FACTORY_DATA_WRITE_PROTECT` as it is the one that selects `FPROTECT`.  so we will have fprotect only for mcuboot on this platform as desired. change does not affect other platforms

PR #513, opened 2024-11-25
[nrf toup] getting boot reason on nRF54h20

kg-nordicsemi:
- Use NRF_RESETINFO register to get boot reason

### DIFF
--- a/src/platform/SaveBootReasonDFUSuit.h
+++ b/src/platform/SaveBootReasonDFUSuit.h
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2021 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef NRF_SAVE_BOOT_REASON_H__
+#define NRF_SAVE_BOOT_REASON_H__
+
+inline int SoftwareBootReasonSUIT __attribute__((section(".noinit")));
+
+inline int getSoftwareRebootReasonSUIT()
+{
+    return SoftwareBootReasonSUIT;
+}
+
+inline void setSoftwareRebootReasonSUIT(int reason)
+{
+    SoftwareBootReasonSUIT = reason;
+}
+
+#endif
--- a/src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
+++ b/src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
@@ -52,6 +52,11 @@ const size_t kMaxHeapSize = CONFIG_SRAM_BASE_ADDRESS + KB(CONFIG_SRAM_SIZE) - PO
 
 #endif
 
+#ifdef CONFIG_SOC_SERIES_NRF54HX
+#include <hal/nrf_resetinfo.h>
+#include <platform/SaveBootReasonDFUSuit.h>
+#endif
+
 namespace chip {
 namespace DeviceLayer {
 
@@ -59,9 +64,54 @@ namespace {
 
 BootReasonType DetermineBootReason()
 {
-#ifdef CONFIG_HWINFO
-    uint32_t reason;
 
+#if defined(CONFIG_SOC_SERIES_NRF54HX) || defined(CONFIG_HWINFO)
+    uint32_t reason;
+#endif
+
+#ifdef CONFIG_SOC_SERIES_NRF54HX
+
+    bool isSoftwareBootReasonSUITInitialized = false;
+    if (!isSoftwareBootReasonSUITInitialized)
+    {
+        setSoftwareRebootReasonSUIT(0);
+        isSoftwareBootReasonSUITInitialized = true;
+    }
+
+    reason = nrf_resetinfo_resetreas_global_get(NRF_RESETINFO);
+
+    if (reason == RESETINFO_RESETREAS_GLOBAL_ResetValue)
+    {
+        return BootReasonType::kSoftwareReset;
+    }
+
+    if (reason & RESETINFO_RESETREAS_GLOBAL_RESETPORONLY_Msk)
+    {
+        return BootReasonType::kBrownOutReset;
+    }
+
+    if (reason & RESETINFO_RESETREAS_GLOBAL_DOG_Msk)
+    {
+        return BootReasonType::kHardwareWatchdogReset;
+    }
+
+    if ((reason & (RESETINFO_RESETREAS_GLOBAL_RESETPIN_Msk | RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk)) ==
+        (RESETINFO_RESETREAS_GLOBAL_RESETPIN_Msk | RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk))
+    {
+        return BootReasonType::kPowerOnReboot;
+    }
+
+    if ((reason & (RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk | RESETINFO_RESETREAS_GLOBAL_SECSREQ_Msk)) ==
+        (RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk | RESETINFO_RESETREAS_GLOBAL_SECSREQ_Msk))
+    {
+        if (GetSoftwareRebootReason() == SoftwareRebootReason::kSoftwareUpdate)
+        {
+            return BootReasonType::kSoftwareUpdateCompleted;
+        }
+    }
+#endif
+
+#ifdef CONFIG_HWINFO
     if (hwinfo_get_reset_cause(&reason) != 0)
     {
         return BootReasonType::kUnspecified;
--- a/src/platform/nrfconnect/OTAImageProcessorImpl.cpp
+++ b/src/platform/nrfconnect/OTAImageProcessorImpl.cpp
@@ -46,6 +46,10 @@
 #include <zephyr/logging/log.h>
 #include <zephyr/pm/device.h>
 
+#ifdef CONFIG_SOC_SERIES_NRF54HX
+#include <platform/SaveBootReasonDFUSuit.h>
+#endif
+
 namespace chip {
 namespace {
 #ifdef CONFIG_CHIP_CERTIFICATION_DECLARATION_STORAGE
@@ -182,6 +186,7 @@ CHIP_ERROR OTAImageProcessorImpl::Apply()
                 PlatformMgr().HandleServerShuttingDown();
                 k_msleep(CHIP_DEVICE_CONFIG_SERVER_SHUTDOWN_ACTIONS_SLEEP_MS);
 #ifdef CONFIG_DFU_TARGET_SUIT
+                SetSoftwareRebootReason(SoftwareRebootReason::kSoftwareUpdate);
                 dfu_target_suit_reboot();
 #else
                 Reboot(SoftwareRebootReason::kSoftwareUpdate);
--- a/src/platform/nrfconnect/Reboot.cpp
+++ b/src/platform/nrfconnect/Reboot.cpp
@@ -25,10 +25,12 @@
 #include <hal/nrf_power.h>
 #endif
 
+#include <platform/SaveBootReasonDFUSuit.h>
+
 namespace chip {
 namespace DeviceLayer {
 
-#if defined(CONFIG_ARCH_POSIX) || defined(CONFIG_SOC_SERIES_NRF54HX)
+#if defined(CONFIG_ARCH_POSIX)
 
 void Reboot(SoftwareRebootReason reason)
 {
@@ -40,7 +42,7 @@ SoftwareRebootReason GetSoftwareRebootReason()
     return SoftwareRebootReason::kOther;
 }
 
-#else
+#elif !(defined(CONFIG_SOC_SERIES_NRF54HX))
 
 using RetainedReason = decltype(nrf_power_gpregret_get(NRF_POWER, 0));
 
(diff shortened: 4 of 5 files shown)


src/platform/nrfconnect/Reboot.cpp
```diff
@@ -28,19 +28,7 @@
 namespace chip {
 namespace DeviceLayer {
 
-#if defined(CONFIG_ARCH_POSIX) || defined(CONFIG_SOC_SERIES_NRF54HX)
```
ArekBalysNordic, line 31: Why is this change needed?

src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
```diff
@@ -60,6 +64,32 @@ namespace {
...
+
+#ifdef CONFIG_SOC_SERIES_NRF54HX
+    uint32_t reason = nrf_resetinfo_resetreas_global_get(NRF_RESETINFO);
+
+    if (reason & RESETINFO_RESETREAS_GLOBAL_RESETPORONLY_Msk)
+    {
+        return BootReasonType::kBrownOutReset;
+    }
+
+    if (reason & RESETINFO_RESETREAS_GLOBAL_DOG_Msk)
```
ArekBalysNordic, line 76: What about changing ifs to else ifs?

ArekBalysNordic (changes requested):

src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
```diff
@@ -60,6 +64,32 @@ namespace {
 BootReasonType DetermineBootReason()
 {
 #ifdef CONFIG_HWINFO
+
+#ifdef CONFIG_SOC_SERIES_NRF54HX
+    uint32_t reason = nrf_resetinfo_resetreas_global_get(NRF_RESETINFO);
```
LuDuda, line 69: Any chance to start using `nrf_` specific API, and instead use HW_INFO module or similar? Probably new nrfx which will be part of NCS 2.9.0 introduces support for nRF54H20 and reset reason.

LuDuda (changes requested):

src/platform/nrfconnect/Reboot.cpp
kg-nordicsemi: Because in nrf54h20 case we don't use Reboot function right? we use dfu_target_suit_reboot. Reset reason is saved in RESETINFO register, so we also don't need to use GetSoftwareRebootReason(), so I thought it is unnecessary to define these functions in case of using nrf54h20.

src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
Damian-Nordic: I know this is subjective, but this is actually opposite to what `clang-tidy` considers readability improvement: https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html :)

src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
```diff
@@ -60,6 +64,32 @@ namespace {
 BootReasonType DetermineBootReason()
 {
 #ifdef CONFIG_HWINFO
+
+#ifdef CONFIG_SOC_SERIES_NRF54HX
```
Damian-Nordic, line 68: Looks like this code doesn't depend on HWINFO, so why is it inside `#ifdef CONFIG_HWINFO`?

src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
```diff
@@ -60,6 +64,32 @@ namespace {
...
+        (RESETINFO_RESETREAS_GLOBAL_RESETPIN_Msk | RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk))
+    {
+        return BootReasonType::kPowerOnReboot;
+    }
+
+    if ((reason & (RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk | RESETINFO_RESETREAS_GLOBAL_SECSREQ_Msk)) ==
+        (RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk | RESETINFO_RESETREAS_GLOBAL_SECSREQ_Msk))
+    {
+        return BootReasonType::kSoftwareUpdateCompleted;
+    }
```
Damian-Nordic, line 91: Don't you need to clear the reset cause somewhere so that it is not accumulated in subsequent reboots?

src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
ArekBalysNordic: Do we know when or how it is planned to be supported?

src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
```diff
@@ -52,16 +52,48 @@ const size_t kMaxHeapSize = CONFIG_SRAM_BASE_ADDRESS + KB(CONFIG_SRAM_SIZE) - PO
...
+    if ((reason & (RESETINFO_RESETREAS_GLOBAL_RESETPIN_Msk | RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk)) ==
+        (RESETINFO_RESETREAS_GLOBAL_RESETPIN_Msk | RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk))
+    {
+        return BootReasonType::kPowerOnReboot;
+    }
 
+    if ((reason & (RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk | RESETINFO_RESETREAS_GLOBAL_SECSREQ_Msk)) ==
+        (RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk | RESETINFO_RESETREAS_GLOBAL_SECSREQ_Msk))
+    {
+        return BootReasonType::kSoftwareUpdateCompleted;
```
LuDuda, line 92: Hmm, how do we know it was caused by the SoftwareUpdate, and not for example crash, or reset from the application layer?

kg-nordicsemi: I performed several manual tests, and checked what are the values returned from RESETINFO in different cases, for example:

- reset after DFU over Matter -> returned value: 528
- reset using button on board -> returned value:514
- reset from the application -> returned value: 0
- reset from disabling power supply -> returned value: 1
                                                                                                             
Events has unique values, and thanks to it, it can be parsed to actual boot reason
But, I don't know what is returning when board is crashing

LuDuda: Right, but I `RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk | RESETINFO_RESETREAS_GLOBAL_SECSREQ_Msk` does not seem to mean that Firmware Upgrade has been done - and the event to Matter core might be generate incorrectly.

We need to either find out if `suit` module might indicate it, or find some register/retention RAM to store information about reset being scheduled due to image update, not for other reason.

@SylwesterKonczyk @tomchy maybe might help if we have anything like this in SUIT, else we need to look at different way.

src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
```diff
@@ -52,16 +52,53 @@ const size_t kMaxHeapSize = CONFIG_SRAM_BASE_ADDRESS + KB(CONFIG_SRAM_SIZE) - PO
...
 
 BootReasonType DetermineBootReason()
 {
-#ifdef CONFIG_HWINFO
+#if defined(CONFIG_SOC_SERIES_NRF54HX) || defined(CONFIG_HWINFO)
     uint32_t reason;
+#endif
+
+#ifdef CONFIG_SOC_SERIES_NRF54HX
+    reason = nrf_resetinfo_resetreas_global_get(NRF_RESETINFO);
```
LuDuda, line 71: I would like to not use this API as it is HW specific.

What about to wait for this PR: https://github.com/nrfconnect/sdk-zephyr/pull/2293 as dependency. And use the API of HW_INFO instead?

src/platform/SaveBootReasonDFUSuit.h
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2021 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef NRF_SAVE_BOOT_REASON_H__
+#define NRF_SAVE_BOOT_REASON_H__
+
+extern int BootReason;
```
ArekBalysNordic, line 10: Why did you do it in this way?

src/platform/nrfconnect/OTAImageProcessorImpl.cpp
```diff
@@ -182,6 +185,8 @@ CHIP_ERROR OTAImageProcessorImpl::Apply()
                 PlatformMgr().HandleServerShuttingDown();
                 k_msleep(CHIP_DEVICE_CONFIG_SERVER_SHUTDOWN_ACTIONS_SLEEP_MS);
 #ifdef CONFIG_DFU_TARGET_SUIT
+                BootReason = 5;
+                ChipLogDetail(DeviceLayer, "Software Boot reason %d", BootReason);
```
ArekBalysNordic, line 189: Debug log?

src/platform/nrfconnect/OTAImageProcessorImpl.cpp
```diff
@@ -182,6 +185,8 @@ CHIP_ERROR OTAImageProcessorImpl::Apply()
                 PlatformMgr().HandleServerShuttingDown();
                 k_msleep(CHIP_DEVICE_CONFIG_SERVER_SHUTDOWN_ACTIONS_SLEEP_MS);
 #ifdef CONFIG_DFU_TARGET_SUIT
+                BootReason = 5;
```
ArekBalysNordic, line 188: Can we use a pre-defined value for this?

src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
```diff
@@ -52,16 +52,60 @@ const size_t kMaxHeapSize = CONFIG_SRAM_BASE_ADDRESS + KB(CONFIG_SRAM_SIZE) - PO
...
+    if ((reason & (RESETINFO_RESETREAS_GLOBAL_RESETPIN_Msk | RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk)) ==
+        (RESETINFO_RESETREAS_GLOBAL_RESETPIN_Msk | RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk))
+    {
+        return BootReasonType::kPowerOnReboot;
+    }
 
+    if ((reason & (RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk | RESETINFO_RESETREAS_GLOBAL_SECSREQ_Msk)) ==
+        (RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk | RESETINFO_RESETREAS_GLOBAL_SECSREQ_Msk))
+    {
+        if (BootReason == 5)
```
ArekBalysNordic, line 100: Please use a pre-defined value here.

src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
```diff
@@ -52,16 +52,60 @@ const size_t kMaxHeapSize = CONFIG_SRAM_BASE_ADDRESS + KB(CONFIG_SRAM_SIZE) - PO
...
 namespace {
 
 BootReasonType DetermineBootReason()
 {
-#ifdef CONFIG_HWINFO
+
+#if defined(CONFIG_SOC_SERIES_NRF54HX) || defined(CONFIG_HWINFO)
     uint32_t reason;
+#endif
+    ChipLogDetail(DeviceLayer, "Software Boot reason %d", BootReason);
```
ArekBalysNordic, line 72: Debug log?

ArekBalysNordic (changes requested):

src/platform/Zephyr/DiagnosticDataProviderImpl.cpp
kg-nordicsemi: sorry I forgot to remove Debug logs

src/platform/SaveBootReasonDFUSuit.h
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2021 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef NRF_SAVE_BOOT_REASON_H__
+#define NRF_SAVE_BOOT_REASON_H__
+
+inline int SoftwareBootReasonSUIT __attribute__((section(".noinit")));
```
LuDuda, line 10: Please move it inside the Reboot.cpp file.

Btw. We can't have a Nordic sepecific file in the `src/platform` file anyway.

Btw2. Do we need to add a `SUIT` in the name?

src/platform/nrfconnect/OTAImageProcessorImpl.cpp
```diff
@@ -182,6 +186,7 @@ CHIP_ERROR OTAImageProcessorImpl::Apply()
                 PlatformMgr().HandleServerShuttingDown();
                 k_msleep(CHIP_DEVICE_CONFIG_SERVER_SHUTDOWN_ACTIONS_SLEEP_MS);
 #ifdef CONFIG_DFU_TARGET_SUIT
+                SetSoftwareRebootReason(SoftwareRebootReason::kSoftwareUpdate);
                 dfu_target_suit_reboot();
```
LuDuda, line 190: Just wonder.. Maybe we can call `Reboot(SoftwareRebootReason::kSoftwareUpdate);` and inside Reboot.cpp call `dfu_target_suit_reboot();` for `kSoftwareUpdate`?

@ArekBalysNordic what do you think?

LuDuda (changes requested):

PR #522, opened 2024-12-06
[nrf toup] Clear Thread Network after last fabric is removed

adigie:
Dataset has to be cleared to allow commissioning to different network when `CONFIG_CHIP_LAST_FABRIC_REMOVED_ERASE_ONLY` is used.

### DIFF
--- a/src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.h
+++ b/src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.h
@@ -87,6 +87,8 @@ public:
         connectNetworkTimeout     = connectTimeoutSec;
     }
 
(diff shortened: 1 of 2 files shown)


doublemis1:
It should be set as a backport (PR to nrf-sdk)

ArekBalysNordic (approved):
Looks good and indeed missing there if we don't want to reset the device and clear RAM. In normal cases, the clearing process takes place only while commissioning in `HandleRemoveNetwork` right?

ArekBalysNordic:
> It should be set as a backport (PR to nrf-sdk)

I guess we need to do it manually.

LuDuda (approved):

maciejbaczmanski (approved):

PR #527, opened 2024-12-20
[nrf toup] Disable mpsl before performing factory data

ArekBalysNordic:
We can speed up flash operations while performing a factory reset by disabling mpsl before that.

Thanks to that flash operations do not wait for mpsl synchronization and all write operations take less time.

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig.defaults
+++ b/config/nrfconnect/chip-module/Kconfig.defaults
@@ -72,6 +72,21 @@ config NVS_LOOKUP_CACHE_SIZE
 config ZMS_LOOKUP_CACHE_SIZE
     default 512 if ZMS
 
+# Enable Dynamic interrupts to have a possibility to reconfigure it for devices that use MPSL
+# on the application core and needs synchronization between MPSL and FLASH driver.
+if MPSL
+
+config DYNAMIC_INTERRUPTS
+    default y
+
+config DYNAMIC_DIRECT_INTERRUPTS
+    default y
+
+config MPSL_DYNAMIC_INTERRUPTS
+    default y
+
+endif # MPSL
+
 # ==============================================================================
 # Zephyr networking configuration
 # ==============================================================================
--- a/src/platform/Zephyr/ConfigurationManagerImpl.cpp
+++ b/src/platform/Zephyr/ConfigurationManagerImpl.cpp
@@ -54,6 +54,10 @@
 #include <platform/ThreadStackManager.h>
 #endif
 
+#ifdef CONFIG_SOC_FLASH_NRF_RADIO_SYNC_MPSL
+#include <mpsl/mpsl_lib.h>
+#endif // CONFIG_SOC_FLASH_NRF_RADIO_SYNC_MPSL
+
 namespace chip {
 namespace DeviceLayer {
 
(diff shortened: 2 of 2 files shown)


src/platform/Zephyr/ConfigurationManagerImpl.cpp
```diff
@@ -54,6 +54,10 @@
 #include <platform/ThreadStackManager.h>
 #endif
 
+#if CHIP_DEVICE_LAYER_TARGET_NRFCONNECT
+#include <mpsl/mpsl_lib.h>
```
doublemis1, line 58: is the mpsl is a common files for all platforms based on zephyr?
or it should be defined in nrfconnect platform?

kkasperczyk-no: +1, mpsl is nordic specific, so ideally we should not do that for Zephyr, otherwise there may be problems with contribution to upstream.

ArekBalysNordic: It is Nordic-specific hence `CHIP_DEVICE_LAYER_TARGET_NRFCONNECT`. Otherwise, we need to implement the whole factory reset method in the nrfconnect platform, because currently, we use the Zephyr one from here. Do you think it is better to move it there?

kkasperczyk-no (approved):

adigie (approved):

src/platform/Zephyr/ConfigurationManagerImpl.cpp
```diff
@@ -220,14 +225,20 @@ void ConfigurationManagerImpl::DoFactoryReset(intptr_t arg)
     }
 #else
 
+    ConnectivityMgr().ErasePersistentInfo();
+    //! Workaround for NCS issue #19795
```
LuDuda, line 229: Why this is workaround, isn't just optimization. The higher erase time with MPSL is justified, by disabling MPSL we do have direct access to flash, thus it's faster. What do you think?

ArekBalysNordic: Hmm, ok that's right, I was thinking about that as a workaround for the time penalty that mpsl causes. But ok, in this context it is optimization. :)

ArekBalysNordic: I've removed "workaround".

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -72,6 +72,21 @@ config NVS_LOOKUP_CACHE_SIZE
...
+# on the application core and needs synchronization between MPSL and FLASH driver.
+if MPSL
+
+config DYNAMIC_INTERRUPTS
+    default y
+
+config DYNAMIC_DIRECT_INTERRUPTS
+    default y
+
+config MPSL_DYNAMIC_INTERRUPTS
```
LuDuda, line 85: We still need it even though we do not re-enable mpsl, right?

LuDuda (approved):

config/nrfconnect/chip-module/Kconfig.defaults
ArekBalysNordic: That's right, we need it anyway if we want to disable mpsl.

PR #539, opened 2025-01-24
[nrfconnect] Add platform crypto for KMU usage

ArekBalysNordic:
Let's review the code here first, before pushing it to the upstream repo.

See commit messages to see the changes.

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig
+++ b/config/nrfconnect/chip-module/Kconfig
@@ -397,6 +397,37 @@ config CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU_ENCRYPTED
 
 endif # CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
 
+config CHIP_STORE_KEYS_IN_KMU
+	bool "Use KMU to store crypto keys"
+	depends on CRACEN_LIB_KMU
+	select EXPERIMENTAL
+	help
+	  Stores cryptographic materials like keys in the KMU if available on the device.
+	  This option utilizes a custom nRF Connect implementation of PSAKeyAllocator for key
+	  storage. All non-volatile cryptographic materials used by the Matter stack will be
+	  stored in KMU slots.
+
+if CHIP_STORE_KEYS_IN_KMU
+
+config CHIP_KMU_SLOT_RANGE_START
+	int "Starting KMU slot number for Matter crypto materials"
+	default 100
+	help
+	  Defines the starting KMU slot number reserved for storing Matter-related keys.
+	  This configuration does not include the DAC private key. To change DAC private key
+	  location see the CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU_SLOT_ID Kconfig.
+
+config CHIP_KMU_SLOT_RANGE_END
+	int "Last available KMU slot for Matter crypto materials"
+	default 175 if CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU
+	default 180
+	help
+	  Defines the last available KMU slot for storing Matter-related keys.
+	  This configuration does not include the DAC private key. To change DAC private key
+	  location see the CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU_SLOT_ID Kconfig.
+
+endif # CHIP_STORE_KEYS_IN_KMU
+
 config CHIP_PERSISTENT_SUBSCRIPTIONS
 	default n
 	# selecting experimental for this feature since there is an issue with multiple controllers.
--- a/config/nrfconnect/chip-module/Kconfig.features
+++ b/config/nrfconnect/chip-module/Kconfig.features
@@ -288,5 +288,4 @@ config CHIP_LAST_FABRIC_REMOVED_ACTION_DELAY
 	  After removing the last fabric the device will wait for the defined time and then perform
 	  an action chosen by the CHIP_LAST_FABRIC_REMOVED_ACTION option. This schedule will allow for
 	  avoiding race conditions before the device removes non-volatile data.
-
 endif # CHIP
--- a/src/crypto/PSAKeyAllocator.h
+++ b/src/crypto/PSAKeyAllocator.h
@@ -1,0 +1,221 @@
+/*
+ *    Copyright (c) 2025 Project CHIP Authors
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#pragma once
+
+#if CHIP_HAVE_CONFIG_H
+#include <crypto/CryptoBuildConfig.h>
+#endif // CHIP_HAVE_CONFIG_H
+
+#include <lib/core/CHIPError.h>
+#include <lib/core/DataModelTypes.h>
+#include <lib/support/TypeTraits.h>
+
+namespace chip {
+namespace Crypto {
+
+static_assert(PSA_KEY_ID_USER_MIN <= CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE && CHIP_CONFIG_CRYPTO_PSA_KEY_ID_END <= PSA_KEY_ID_USER_MAX,
+              "Matter specific PSA key range doesn't fit within PSA allowed range");
+
+// Each ICD client requires storing two keys- AES and HMAC
+static constexpr uint32_t kMaxICDClientKeys = 2 * CHIP_CONFIG_CRYPTO_PSA_ICD_MAX_CLIENTS;
+
+static_assert(kMaxICDClientKeys >= CHIP_CONFIG_ICD_CLIENTS_SUPPORTED_PER_FABRIC * CHIP_CONFIG_MAX_FABRICS,
+              "Number of allocated ICD key slots is lower than maximum number of supported ICD clients");
+
+/**
+ * @brief Defines subranges of the PSA key identifier space used by Matter.
+ */
+enum class KeyIdBase : psa_key_id_t
+{
+    Minimum          = CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE,
+    Operational      = Minimum, ///< Base of the PSA key ID range for Node Operational Certificate private keys
+    DACPrivKey       = Operational + kMaxValidFabricIndex + 1,
+    ICDKeyRangeStart = DACPrivKey + 1,
+    Maximum          = ICDKeyRangeStart + kMaxICDClientKeys,
+};
+
+static_assert(to_underlying(KeyIdBase::Minimum) >= CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE &&
+                  to_underlying(KeyIdBase::Maximum) <= CHIP_CONFIG_CRYPTO_PSA_KEY_ID_END,
+              "PSA key ID base out of allowed range");
+
+/**
+ * @brief Finds first free persistent Key slot ID within range.
+ *
+ * @param[out] keyId Key ID handler to which free ID will be set.
+ * @param[in]  start Starting ID in search range.
+ * @param[in]  range Search range.
+ *
+ * @retval CHIP_NO_ERROR               On success.
+ * @retval CHIP_ERROR_INTERNAL         On PSA crypto API error.
+ * @retval CHIP_ERROR_NOT_FOUND        On no free Key ID within range.
+ * @retval CHIP_ERROR_INVALID_ARGUMENT On search arguments out of PSA allowed range.
+ */
+CHIP_ERROR FindFreeKeySlotInRange(psa_key_id_t & keyId, psa_key_id_t start, uint32_t range);
+
+/**
+ * @brief Calculates PSA key ID for Node Operational Certificate private key for the given fabric.
+ */
+constexpr psa_key_id_t MakeOperationalKeyId(FabricIndex fabricIndex)
+{
+    return to_underlying(KeyIdBase::Operational) + static_cast<psa_key_id_t>(fabricIndex);
+}
+
+/**
+ * @brief Interface for PSA key allocation.
+ *
+ * The PSA Key Allocator interface provides an abstraction that allows the application to
+ * allocate PSA keys in a secure environment. This class uses a concept that isolates the
+ * application from the actual key material. The secure location may vary depending on the
+ * cryptographic hardware used. Using this class a platform can implement this interface to
+ * allocate keys in the specific secure location.
+ *
+ * In some cases key attributes must be redefined to match the specific requirements of the
+ * secure location and the cryptographic hardware.
+ *
+ * This class keeps the static instance and uses it in the all the places where the key
+ * should be persisted. You can add the usage of this class anywhere in the code by calling the
+ * GetPSAKeyAllocator() function.
+ *
+ * If the static instance is not set, the default implementation is used.
+ *
+ * To change the static instance of the PSAKeyAllocator, you can call the SetPSAKeyAllocator function.
+ */
+class PSAKeyAllocator
+{
+public:
+    /**
+     * @brief Destructor for PSAKeyAllocator.
+     */
+    virtual ~PSAKeyAllocator() = default;
+
+    /**
+     * @brief Get the Device Attestation Key (DAC) ID.
+     *
+     * @return psa_key_id_t The DAC key ID.
+     */
+    virtual psa_key_id_t GetDacKeyId() = 0;
+
+    /**
+     * @brief Get the Node Operational Certificate key ID for a given fabric index.
+     *
+     * @param fabricIndex The fabric index for which the operational key ID is requested.
+     * @return psa_key_id_t The operational key ID.
+     */
+    virtual psa_key_id_t GetOpKeyId(FabricIndex fabricIndex) = 0;
+
+    /**
+     * @brief Allocate a new Intermittently Connected Devices (ICD) key ID.
+     *
+     * This method is used to allocate both AES-CCM and HMAC (SHA-256) keys independently.
+     * The caller is responsible for storing the key ID in non-volatile memory
+     * and setting the appropriate key type.
+     *
+     * @return psa_key_id_t The newly allocated ICD key ID.
+     */
+    virtual psa_key_id_t AllocateICDKeyId() = 0;
+
+    /**
+     * @brief Update the key attributes before storing the key.
+     *
+     * In some cases the key attributes must be redefined to match the specific requirements of the
+     * secure location and the cryptographic hardware. This method allows the platform to update the
+     * key attributes before storing the key.
+     *
+     * Read the current key attributes to determine the key type, algorithm, and usage flags. Update
+     * the key attributes as needed.
+     *
+     * @param attrs Reference to the key attributes structure to be updated.
+     */
+    virtual void UpdateKeyAttributes(psa_key_attributes_t & attrs) = 0;
+
+    // Allow setting and getting the static instance of the PSAKeyAllocator by external functions
+    friend PSAKeyAllocator & GetPSAKeyAllocator();
+    friend void SetPSAKeyAllocator(PSAKeyAllocator * keyAllocator);
+
+private:
+    static PSAKeyAllocator * sInstance;
+};
+
+/**
+ * @brief Default implementation of PSAKeyAllocator.
+ *
+ * This default implementation allocates key IDs according to the KeyIdBase enum.
+ * The operational key ID is calculated as the base operational key ID plus the fabric index.
+ * The DAC key ID is calculated as the base DAC key ID.
+ * The ICD key ID is allocated from the range starting from the ICDKeyRangeStart.
+ * The key attributes are not updated.
+ */
+class DefaultPSAKeyAllocator : public PSAKeyAllocator
+{
+public:
+    // implementations of the PSAKeyAllocator interface
+    psa_key_id_t GetDacKeyId() override { return to_underlying(KeyIdBase::DACPrivKey); }
+    psa_key_id_t GetOpKeyId(FabricIndex fabricIndex) override { return MakeOperationalKeyId(fabricIndex); }
+    psa_key_id_t AllocateICDKeyId() override
+    {
+        psa_key_id_t newKeyId = PSA_KEY_ID_NULL;
+        if (CHIP_NO_ERROR !=
+            Crypto::FindFreeKeySlotInRange(newKeyId, to_underlying(KeyIdBase::ICDKeyRangeStart), kMaxICDClientKeys))
+        {
+            newKeyId = PSA_KEY_ID_NULL;
+        }
+        return newKeyId;
+    }
+    void UpdateKeyAttributes(psa_key_attributes_t & attrs) override
+    {
+        // Do nothing
+    }
+};
+
+/**
+ * @brief Static function to get the instance of PSAKeyAllocator.
+ *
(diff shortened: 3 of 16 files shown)


src/platform/nrfconnect/crypto/KMUSessionKeystore.cpp
```diff
@@ -0,0 +1,202 @@
...
+    return error;
+}
+
+#if CHIP_CONFIG_ENABLE_ICD_CIP
+CHIP_ERROR KMUSessionKeystore::PersistICDKey(Symmetric128BitsKeyHandle & key)
+{
+    CHIP_ERROR err;
+    psa_key_id_t newKeyId = PSA_KEY_ID_NULL;
+    psa_key_attributes_t attrs;
+    psa_status_t s = PSA_SUCCESS;
```
kkasperczyk-no, line 139: Is it used anywhere?

config/nrfconnect/chip-module/Kconfig.features
```diff
@@ -289,4 +289,43 @@ config CHIP_LAST_FABRIC_REMOVED_ACTION_DELAY
...
+	int
+	default 175 if CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU
+	default 180
+	help
+	  The last available KMU slot for Matter purposes.
+	  This slot does not include DAC private key.
+
+	config CHIP_KMU_MAX_FABRICS
+	int
+	default 18 if (CHIP_ENABLE_ICD_SUPPORT && CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU)
```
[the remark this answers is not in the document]
kkasperczyk-no: Also, are you sure we should check `CHIP_ENABLE_ICD_SUPPORT` and not `CHIP_CONFIG_ENABLE_ICD_CIP`? I think that we may use ICD without CIP and in such case we don't need additional keys. Right @maciejbaczmanski ?

src/platform/nrfconnect/crypto/KMUSlotDefinitions.h
```diff
@@ -0,0 +1,52 @@
...
+#error "CONFIG_CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU_SLOT_ID Cannot overlaps with KMU slots dedicated for Matter core"
+#endif
+
+// Define how many slots are available for Matter.
+#define KMU_MATTER_SLOT_COUNT (CONFIG_CHIP_CORE_KMU_SLOT_END - CONFIG_CHIP_CORE_KMU_SLOT_START)
+// For NOC we need 2 slots per fabric (ESDSA)
+#define KMU_MATTER_NOC_SLOT_COUNT (CONFIG_CHIP_KMU_MAX_FABRICS * 2)
+#define KMU_MATTER_SLOTS_PER_NOC_KEY 2
+
+#ifdef CONFIG_CHIP_ENABLE_ICD_SUPPORT
```
kkasperczyk-no, line 33: I would say here it should be CHIP_CONFIG_ENABLE_ICD_CIP instead of `CONFIG_CHIP_ENABLE_ICD_SUPPORT` as well

src/crypto/crypto.gni
```diff
@@ -28,5 +28,5 @@ assert(
     !chip_external_mbedtls || chip_crypto == "mbedtls" || chip_crypto == "psa",
     "Use of external mbedtls requires the mbedtls or psa crypto impl")
 
-assert(!chip_crypto_psa_spake2p || chip_crypto == "psa",
-       "Use of psa spake2+ requires the psa crypto impl")
+assert(!chip_crypto_psa_spake2p || chip_crypto == "psa" || chip_crypto == "platform",
```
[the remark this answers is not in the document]
ArekBalysNordic: No, I just allowed to use psa spake2p within the platform implementation, not require it there. Without this change, it is not allowed to use psa spake2p from crypto if the crypto=platform is selected. This doesn't force anything.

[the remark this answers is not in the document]
ArekBalysNordic: But I agree that change is a little bit controversial, but having here `chip_device_platform == "nrfconnect"` is even more controversial I guess. :)

src/platform/nrfconnect/crypto/KMUSessionKeystore.cpp
ArekBalysNordic: no, it was for testing purposes, will remove it.

src/crypto/crypto.gni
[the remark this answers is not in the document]
Damian-Nordic: The question is if we need to use `chip_crypto = "platform"`. You can  It will build `PSAOperationalKeystore` and other PSA bits but you can override them at runtime anyway.

config/nrfconnect/chip-module/Kconfig.features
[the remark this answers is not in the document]
ArekBalysNordic: I've changed it to calculate it in the header file.

src/crypto/crypto.gni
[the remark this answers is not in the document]
ArekBalysNordic: hmm, that's right. It started to be possible once those classes were inheritable. :)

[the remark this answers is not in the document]
ArekBalysNordic: Ok, great! it looks much better now. :) I will resolve this comment then. Thanks.

src/crypto/PSASessionKeystore.cpp
```diff
@@ -189,9 +190,9 @@ void PSASessionKeystore::DestroyKey(HkdfKeyHandle & key)
 #if CHIP_CONFIG_ENABLE_ICD_CIP
 CHIP_ERROR PSASessionKeystore::PersistICDKey(Symmetric128BitsKeyHandle & key)
 {
-    CHIP_ERROR err;
-    psa_key_id_t newKeyId      = PSA_KEY_ID_NULL;
-    psa_key_attributes_t attrs = PSA_KEY_ATTRIBUTES_INIT;
+    CHIP_ERROR err        = CHIP_NO_ERROR;
+    psa_key_id_t newKeyId = PSA_KEY_ID_NULL;
+    psa_key_attributes_t attrs;
```
adigie, line 195:
```suggestion
    psa_key_attributes_t attrs = PSA_KEY_ATTRIBUTES_INIT;
```

ArekBalysNordic: hmm, something went wrong during the rebase.

Damian-Nordic (commented):
Looks great. Some mostly minor suggestions from me.

src/platform/nrfconnect/KMUKeyAllocator.h
```diff
@@ -0,0 +1,142 @@
...
+ *    limitations under the License.
+ */
+#pragma once
+#include <cracen_psa_kmu.h>
+#include <crypto/PSAKeyAllocator.h>
+// Define the number of slots per NOC and ICD key.
+#define KMU_MATTER_SLOTS_PER_NOC_KEY 2
+#define KMU_MATTER_SLOTS_PER_ICD_KEY 2
+// Define the maximum number of fabrics supported by the KMU.
+#if CONFIG_CHIP_ICD_CHECK_IN_SUPPORT
```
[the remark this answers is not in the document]
ArekBalysNordic: I'll discuss it with the team shortly if we even plan such a case.

src/platform/nrfconnect/KMUKeyAllocator.h
```diff
@@ -0,0 +1,142 @@
...
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+#pragma once
+#include <cracen_psa_kmu.h>
+#include <crypto/PSAKeyAllocator.h>
+// Define the number of slots per NOC and ICD key.
+#define KMU_MATTER_SLOTS_PER_NOC_KEY 2
+#define KMU_MATTER_SLOTS_PER_ICD_KEY 2
```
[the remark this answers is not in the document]
ArekBalysNordic: That's right, but it is consistent with NOC :) I will add a comment there.

src/platform/nrfconnect/KMUKeyAllocator.h
```diff
@@ -0,0 +1,142 @@
...
+// Check whether the DAC KMU slot doesn not overlap with the KMU slots dedicated for Matter core.
+#if defined(CONFIG_CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU) &&                                                                            \
+    (CONFIG_CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU_SLOT_ID >= CONFIG_CHIP_CORE_KMU_SLOT_START &&                                         \
+     CONFIG_CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU_SLOT_ID <= CONFIG_CHIP_CORE_KMU_SLOT_END)
+#error "CONFIG_CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU_SLOT_ID Cannot overlaps with KMU slots dedicated for Matter core"
+#endif
+// Check if the maximum of used slots is not greater than the available slots.
+// Each fabric requires 2 slots (ESDSA), one ICD entry requires 2 slots (AES + HMAC).
+// If migration DAC to KMU is enabled we need to reserve slots for it.
+#if (KMU_MATTER_NOC_SLOT_COUNT + KMU_MATTER_ICD_SLOT_COUNT) > KMU_MATTER_SLOT_COUNT
```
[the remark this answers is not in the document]
ArekBalysNordic: yeah, it is redundant now, after the recent changes.

src/platform/nrfconnect/KMUKeyAllocator.h
```diff
@@ -0,0 +1,142 @@
...
+            psa_set_key_algorithm(attrs, PSA_ALG_ECDSA(PSA_ALG_ANY_HASH));
+        }
+
+        // TODO: Change PSA_ALG_AEAD_WITH_AT_LEAST_THIS_LENGTH_TAG(PSA_ALG_CCM, 8) to PSA_ALG_CCM
+        // To be compatible with the KMU. Remove this workaround once KMU supports PSA_ALG_AEAD_WITH_AT_LEAST_THIS_LENGTH_TAG.
+        // Currently we need to use EXPORT flag because the Cracen does not support copying keys from ITS to KMU
+        if (psa_get_key_algorithm(attrs) == PSA_ALG_AEAD_WITH_AT_LEAST_THIS_LENGTH_TAG(PSA_ALG_CCM, 8))
+        {
+            psa_set_key_algorithm(attrs, PSA_ALG_CCM);
+            psa_set_key_usage_flags(attrs, psa_get_key_usage_flags(attrs) | PSA_KEY_USAGE_EXPORT);
```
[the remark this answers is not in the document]
ArekBalysNordic: Even if it is supported in KMU, indeed we should disable it while storing it there :)

src/platform/nrfconnect/KMUKeyAllocator.h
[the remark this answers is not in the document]
ArekBalysNordic: @kkasperczyk-no @LuDuda what do you think?

config/nrfconnect/chip-module/Kconfig.features
```diff
@@ -289,4 +289,30 @@ config CHIP_LAST_FABRIC_REMOVED_ACTION_DELAY
 	  an action chosen by the CHIP_LAST_FABRIC_REMOVED_ACTION option. This schedule will allow for
 	  avoiding race conditions before the device removes non-volatile data.
 
+config CHIP_STORE_KEYS_IN_KMU
```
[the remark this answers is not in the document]
ArekBalysNordic: Moved.

src/platform/nrfconnect/KMUSessionKeystore.h
```diff
@@ -0,0 +1,76 @@
...
+    {
+        CHIP_ERROR err        = CHIP_NO_ERROR;
+        psa_key_id_t newKeyId = PSA_KEY_ID_NULL;
+        psa_key_attributes_t attrs;
+        uint8_t keyEx[chip::Crypto::CHIP_CRYPTO_SYMMETRIC_KEY_LENGTH_BYTES];
+        size_t keyExSize = sizeof(keyEx);
+
+        psa_get_key_attributes(key.As<psa_key_id_t>(), &attrs);
+
+        if (psa_get_key_lifetime(&attrs) == PSA_KEY_LIFETIME_PERSISTENT)
```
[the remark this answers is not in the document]
ArekBalysNordic: Right, it should be check related to KMU location :)

Damian-Nordic (approved):
Thanks for the changes to reduce the duplicated code! Approving assuming you will be discussing the non-ICD to ICD upgrade in the team.

config/nrfconnect/chip-module/Kconfig
```diff
@@ -397,6 +397,33 @@ config CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU_ENCRYPTED
 
 endif # CHIP_CRYPTO_PSA_MIGRATE_DAC_PRIV_KEY
 
+config CHIP_STORE_KEYS_IN_KMU
+	bool "Use KMU to store crypto keys"
+	default y
```
LuDuda, line 402: I think this should be default configuration at some point of time, not sure if we should enable it right away without thorough testing. Also, for now, i would limit it to nRF54L, not sure if H won't enable automatically?

config/nrfconnect/chip-module/Kconfig
```diff
@@ -397,6 +397,33 @@ config CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU_ENCRYPTED
...
+	depends on CRACEN_LIB_KMU
+	help
+	  Uses KMU to store keys if it is available on the device. This option uses
+	  nrfconnect custom crypto implementation of storing keys. All crypto materials
+	  used in Matter stack will be persisted in the KMU slots.
+
+if CHIP_STORE_KEYS_IN_KMU
+	config CHIP_CORE_KMU_SLOT_START
```
LuDuda, line 410: nit: I think we do not use alignments for KConfigs (right?) plus add a new line above

config/nrfconnect/chip-module/Kconfig
```diff
@@ -397,6 +397,33 @@ config CHIP_CRYPTO_PSA_DAC_PRIV_KEY_KMU_ENCRYPTED
...
+	int "Initial KMU slot number for Matter keys"
+	default 100
+	help
+	  Starting slot number dedicated for Matter purposes.
```
LuDuda, line 414: Maybe: `Defines the initial KMU slot number reserved for storing Matter-related keys. This configuration does not include the DAC private key.`

src/crypto/PSAKeyAllocator.h
```diff
@@ -0,0 +1,212 @@
...
+     * secure location and the cryptographic hardware. This method allows the platform to update the
+     * key attributes before storing the key.
+     *
+     * Read the current key attributes to determine the key type, algorithm, and usage flags. Update
+     * the key attributes as needed.
+     *
+     * @param attrs Pointer to the key attributes structure to be updated.
+     */
+    virtual void UpdateKeyAttributes(psa_key_attributes_t & attrs) = 0;
+
```
LuDuda, line 142: Out of scope of this PR: I do understand that Group Keys are not stored in non-volatile secure storage currently?

src/crypto/PSAKeyAllocator.h
```diff
@@ -0,0 +1,212 @@
...
+     * In some cases the key attributes must be redefined to match the specific requirements of the
+     * secure location and the cryptographic hardware. This method allows the platform to update the
+     * key attributes before storing the key.
+     *
+     * Read the current key attributes to determine the key type, algorithm, and usage flags. Update
+     * the key attributes as needed.
+     *
+     * @param attrs Pointer to the key attributes structure to be updated.
+     */
+    virtual void UpdateKeyAttributes(psa_key_attributes_t & attrs) = 0;
```
LuDuda, line 141: Not a use-case for now... but should we add actually the `psa_key_id_t` as input parameter? Let's imagine that we would like to somehom differently treat some of the keys like DAC in the future (or even now).

src/crypto/PSAKeyAllocator.h
```diff
@@ -0,0 +1,212 @@
...
+     * @param fabricIndex The fabric index for which the operational key ID is requested.
+     * @return psa_key_id_t The operational key ID.
+     */
+    virtual psa_key_id_t GetOpKeyId(FabricIndex fabricIndex) = 0;
+
+    /**
+     * @brief Allocate a new Intermittently Connected Devices (ICD) key ID.
+     *
+     * ICD uses two keys - AES CCM and HMAC(SHA 256).
+     * Each key must be allocated separately.
```
LuDuda, line 123: So what is the key_id_t returned here? We are making assumption that HMAC(SHA256) is always returned value + 1? Maybe, we should return some structure here actually?

Just a though from scalability perspective @ArekBalysNordic  @Damian-Nordic ?

src/platform/nrfconnect/KMUKeyAllocator.h
```diff
@@ -0,0 +1,136 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+#pragma once
+#include <cracen_psa_kmu.h>
+#include <crypto/PSAKeyAllocator.h>
+// Define the number of slots per NOC and ICD key.
```
LuDuda, line 20: nit: but can we please make some more new lines in between of blocks for better readability

src/platform/nrfconnect/KMUKeyAllocator.h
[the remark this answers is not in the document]
LuDuda: We should somehow allow customer for upgradability (which of course is harder with KMU), even if customer will need to modify manually KMUKeyAllocator implementation to allocate keys from outside of the Matter range . We need to describe it in the documentation thoroughly. If i correctly reading the code, someone would be able to modify the code for it's own purpose and manage key_ids as he wants, right? Even for example to store ICD keys to ITS instead of KMUs (with known side-effects)?

Not sure if we should always blindly reserve slots which will not be used.. Maybe we need F2F discussion. Is it possible to "move" keys from slot A to slot B?

src/crypto/PSAKeyAllocator.h
LuDuda: Update: Oh! Now i see below, that it's possible to read `psa_key_id` from `psa_key_attributes_t` :)

src/platform/nrfconnect/KMUKeyAllocator.h
```diff
@@ -0,0 +1,136 @@
...
+        if (psa_get_key_id(&attrs) >=
+                PSA_KEY_HANDLE_FROM_CRACEN_KMU_SLOT(CRACEN_KMU_KEY_USAGE_SCHEME_RAW, CONFIG_CHIP_CORE_KMU_SLOT_START) &&
+            psa_get_key_id(&attrs) <
+                PSA_KEY_HANDLE_FROM_CRACEN_KMU_SLOT(CRACEN_KMU_KEY_USAGE_SCHEME_RAW, CONFIG_CHIP_CORE_KMU_SLOT_END))
+        {
+            psa_set_key_lifetime(
+                &attrs, PSA_KEY_LIFETIME_FROM_PERSISTENCE_AND_LOCATION(PSA_KEY_PERSISTENCE_DEFAULT, PSA_KEY_LOCATION_CRACEN_KMU));
+        }
+        // TODO: Currently ECDSA keys are supported only if HASH_ANY is provided, so change it.
+        // Remove this workaround once KMU supports ECSA SHA256.
```
LuDuda, line 92:
```suggestion
        // Remove this workaround once KMU supports ECDSA SHA256.
```

LuDuda (commented):
This is very clean! Great work 👍 

Few nits and questions below.

config/nrfconnect/chip-module/Kconfig
ArekBalysNordic: There is no KMU in H. :) This feature is limited to nRF54L only.

src/crypto/PSAKeyAllocator.h
ArekBalysNordic: that's right, actually, in this way, it is clearer to me that we're going to edit attributes.

src/crypto/PSAKeyAllocator.h
ArekBalysNordic: AFAIK, they're stored in settings.

config/nrfconnect/chip-module/Kconfig
ArekBalysNordic: I will remove `default y` before merging it.

config/nrfconnect/chip-module/Kconfig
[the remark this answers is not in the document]
ArekBalysNordic: OK, "RANGE" looks better. :) and "CORE" was used to exclude the DAC private key, but it is better to describe it in help, instead of naming it "CORE". So I'll remove it. :)

src/crypto/PSAKeyAllocator.h
LuDuda: Let's discuss it F2F - not sure I understand how it works then if we want to make operation HMAC(SHA256). On which Key ID we do this? Even if it's stored in settings, how we get it's number?

LuDuda (approved):
Looks good for me too. Let's discuss F2F the approach for upgreadability.

src/crypto/PSAKeyAllocator.h
ArekBalysNordic: During the ICD registration, we create two keys - HMAC(SHA256) and AES CCM. Originally, we put them into RAM and save their key IDs to Zephyr settings. If we need to use them, we get the key id from the specific Zephyr settings entry (used for ICD purposes) and provide it to PSA API. Recently we added moving these keys to PSA ITS (Now to KMU). Now during the ICD registration, we call an additional function "PersistICDKey" on each key and thanks to that we move the keys from RAM to ITS/KMU and we update the key ID saved in settings. So, to sum up, in settings we have something like a look-up table and key IDs assigned to specific ICD settings entries. Therefore, we don't need to take care of a specific slot for those keys, because we already have this information saved in zephyr settings. :)

LuDuda: This part i get, but i don't understand what is the Key ID for AES CCM and what is the Key ID for HMAC(SHA256)? Looking at the code, I assume the `AllocateICDKeyId` returns Key ID for AES CCM slot only - so what happens with HMAC(SHA256)? Or this is not needed to be persist?

Damian-Nordic: `AllocateICDKeyId` just allocates a key ID from a given range. The caller then uses the key ID to either create AES or HMAC key. It then stores in setting that for a given ICD entry, AES key ID is, say, 0x27 and HMAC key ID is, say, 0x32.

ArekBalysNordic: Fortunately, both AES and HMAC costs 1 KMU slot, so we can do it in the same way as for ITS. During the registration both keys are created one by one and uses the same allocate method. So Aes is created, allocate metod returns the first available key id within the given range, then HMAC is created and allocate method returns the next available key id in the same range. The difference is during the key creation. We use different attributes for each key, and then we don't need do check the key type during the allocation, just allocate the following ids.

LuDuda: Oh, right.. Somehow i thought that during allocation of the slot, we also do need to set a proper `key_type`, but it's more about just get an ID from a range. I think it works then fine.

LuDuda: Still wound't be more scalable to return struct with two key_ids already? As you mentioned it just happened that they both use 1 slot of KMU, what if this change for any reason in future? Else, i would expand the documentation to make it a bit more clear, maybe:

```
 * This method is used to allocate both AES-CCM and HMAC (SHA-256) keys independently.
 * The caller is responsible for storing the key ID in non-volatile memory
 * and setting the appropriate key type.
 ```

ArekBalysNordic: I will change the documentation then, and we can consider changing it to struct as a follow-up because it will require some changes in the Matter stack.

PR #540, opened 2025-01-28
[crypto] Fix initialization of psa key attributes

adigie:
psa_key_attributes_t has to be initialized with PSA_KEY_ATTRIBUTES_INIT.

### DIFF
--- a/src/crypto/PSASessionKeystore.cpp
+++ b/src/crypto/PSASessionKeystore.cpp
@@ -190,8 +190,8 @@ void PSASessionKeystore::DestroyKey(HkdfKeyHandle & key)
 CHIP_ERROR PSASessionKeystore::PersistICDKey(Symmetric128BitsKeyHandle & key)
 {
     CHIP_ERROR err;
-    psa_key_id_t newKeyId = PSA_KEY_ID_NULL;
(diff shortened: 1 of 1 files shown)


ArekBalysNordic (approved):

LuDuda:
@adigie what are the consequences of previous behaviour?

Damian-Nordic:
This seems unnecessary according to https://arm-software.github.io/psa-api/crypto/1.1/api/keys/attributes.html#c.psa_get_key_attributes
> This function first resets the attribute object as with [psa_reset_key_attributes()](https://arm-software.github.io/psa-api/crypto/1.1/api/keys/attributes.html#c.psa_reset_key_attributes). It then copies the attributes of the given key into the given attribute object.

adigie:
It freed some random address in the upstream tests on 2.7.0, because of:
https://github.com/nrfconnect/sdk-mbedtls/blob/72868c6f1421afa92cba9f6fe6d1a3f108c174b7/library/psa_crypto_client.c#L19-L19

Now I see that we don't have domain parameters anymore:
https://github.com/nrfconnect/sdk-mbedtls/blob/HEAD/library/psa_crypto_client.c#L19-L19

PR #547, opened 2025-01-31
[nrf noup] Add zcl_generate to west command

ArekBalysNordic:
Added zap_append command to the west.

use `west zap-append -h` to see all options.

Now you can add new clusters to the matter ZCL datamodel.
Use a `west zap-append` and provide the new cluster definitionsas XML files to add them to the zcl.json file.
You can generate the new zcl.json file by providing `-o, --output` argument, or overwrite the existing one.

The zcl_generate is integrated with `west zap-gui` commands.
If you want to use it call the `west zap-gui` command with the additional argument `--clusters`.
You can provide multiple clusters at once.
If you provide an additional `-j` / `--zcl-json` argument alongside the `--clusters` argument to the `west zap-gui` command, the new zcl.json will be created and contain the new clusters.

Tested on MacOS with Intel, and Ubuntu 24.04, and Windows 11

manifest-pr-skip

### DIFF
--- a/scripts/west/west-commands.yml
+++ b/scripts/west/west-commands.yml
@@ -10,3 +10,8 @@ west-commands:
       - name: zap-gui
         class: ZapGui
         help: Run Matter ZCL Advanced Platform (ZAP) GUI
+  - file: scripts/west/zap_append.py
+    commands:
+      - name: zap-append
+        class: ZapAppend
+        help: Append a new custom cluster to the ZCL database
--- a/scripts/west/zap_append.py
+++ b/scripts/west/zap_append.py
@@ -1,0 +1,108 @@
+# Copyright (c) 2025 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+
+import argparse
+from pathlib import Path
+import json
+
+from west.commands import WestCommand
+from west import log
+
+from zap_common import DEFAULT_MATTER_PATH, DEFAULT_ZCL_JSON_RELATIVE_PATH
+
+
+def add_cluster_to_zcl(zcl_base: Path, cluster_xml_paths: list, output: Path):
+    """
+    Add the cluster to the ZCL file.
+    """
+
+    try:
+        with open(zcl_base, "r") as zcl_json_base:
+            zcl_json = json.load(zcl_json_base)
+    except IOError as e:
+        raise RuntimeError(f"No such ZCL file: {zcl_base}")
+
+    # If the output file is provided, we would like to generate a new ZCL file, so we must set
+    # the relative paths from the xml.file to the data model directories and manufacturers.xml file.
+    # It is because the base zcl.json file contains the relative paths to itself inside as the "xmlRoot",
+    # and if we create a new zcl file outside the data model directory it will not work properly.
+    if output:
+        roots_replaced = list()
+        replace = False
+
+        # Replace existing paths with the relative to output ones
+        for path in zcl_json.get("xmlRoot"):
+            path = zcl_base.parent.joinpath(Path(path))
+            if not path == "./" and not path == "." and not path.is_relative_to(output.parent):
+                roots_replaced.append(str(path.relative_to(output.parent, walk_up=True)))
+                replace = True
+        if replace:
+            zcl_json["xmlRoot"] = roots_replaced
+
+        # Add the relative path to manufacturers XML
+        manufacturers_xml = zcl_base.parent.joinpath(Path(zcl_json.get("manufacturersXml"))).resolve()
+        if not manufacturers_xml.parent.is_relative_to(output.parent):
+            zcl_json.update({"manufacturersXml": str(manufacturers_xml.relative_to(output.parent, walk_up=True))})
+
+    # Add the new clusters to the ZCL file
+    for cluster in cluster_xml_paths:
+        if not Path(cluster).exists():
+            raise RuntimeError(f"No such cluster file: {cluster}")
+
+        # Get cluster file name
+        file = Path(cluster).name
+        # Get relative path from the cluster file to the output file.
+        relative_path = Path(cluster).absolute().parent.relative_to(output.parent, walk_up=True)
+
+        # We need to add two things:
+        # 1. The absolute path to the directory where the new xml file exists to the xmlRoot array.
+        # 2. The new xml file name to the xmlFile array.
+        if not str(relative_path) in zcl_json.get("xmlRoot"):
+            zcl_json.get("xmlRoot").append(str(relative_path))
+        if not file in zcl_json.get("xmlFile"):
+            zcl_json.get("xmlFile").append(file)
+            log.dbg(f"Successfully added {file}")
+
+    # If output file is not provided, we will edit the existing ZCL file
+    file_to_write = output if output else zcl_base
+
+    # Save the dumped JSON to the output file
+    with open(file_to_write, "w+") as zcl_output:
+        zcl_output.write(json.dumps(zcl_json, indent=4))
+
+
+class ZapAppend(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'zap-append',
+            'Add a new custom cluster to the ZCL Matter data model file',
+            'A tool for adding a custom cluster to the ZCL Matter Data Model file according to the base ZCL file and custom clusters definitions.')
+
+    def do_add_parser(self, parser_adder) -> argparse.ArgumentParser:
+        parser = parser_adder.add_parser(self.name,
+                                         help=self.help,
+                                         formatter_class=argparse.RawDescriptionHelpFormatter,
+                                         description=self.description)
+        parser.add_argument("-b", "--base", type=Path,
+                            help=f"An absolute path to the base zcl.json file. If not provided the path will be set to MATTER/{DEFAULT_ZCL_JSON_RELATIVE_PATH}.")
+        parser.add_argument("-m", "--matter", type=Path, default=DEFAULT_MATTER_PATH,
+                            help=f"An absolute path to the Matter directory. If not set the path with be set to the {DEFAULT_MATTER_PATH}")
+        parser.add_argument("-o", "--output", type=Path,
+                            help=f"Output path to store the generated zcl.json file. If not provided the path will be set to the base zcl.json file (MATTER/{DEFAULT_ZCL_JSON_RELATIVE_PATH}).")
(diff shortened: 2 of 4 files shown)


scripts/west/west-commands.yml
```diff
@@ -10,3 +10,8 @@ west-commands:
       - name: zap-gui
         class: ZapGui
         help: Run Matter ZCL Advanced Platform (ZAP) GUI
+  - file: scripts/west/zcl_generate.py
+    commands:
+      - name: zcl-generate
+        class: ZCLGenerate
+        help: Generate ZCL code from XML cluster definition file
```
kkasperczyk-no, line 17: I think naming is a bit unfortunate here and there. According to name and help it seems that it generates source code, while it's not true, as it generates only json file that describes data model. Source code is generated by zap-generate, which is also unfortunate name, because it sounds like it could generate zap file, not the code.

I don't know if we should change zap-generate name, as customers could already learn how to use it, but I would try to name the new command in a different way (and change help as well). Maybe zap-generate-json?

ArekBalysNordic: Or maybe `zcl-append-cluster`

kkasperczyk-no: But other commands start with zap, so why we want to use zcl now?

ArekBalysNordic: Because, actually it is more about ZCL than zap.

scripts/west/zap_common.py
```diff
@@ -20,6 +21,9 @@
 from west import log
 
 DEFAULT_MATTER_PATH = Path(__file__).parents[2]
+DEFAULT_DATA_MODEL_PATH = f'{DEFAULT_MATTER_PATH}/src/app/zap-templates/zcl'
+DEFAULT_ZCL_PATH = f'{DEFAULT_DATA_MODEL_PATH}/zcl.json'
+DEFAULT_APP_TEMPLATES_PATH = f'{DEFAULT_MATTER_PATH}/src/app/zap-templates/app-templates.json'
```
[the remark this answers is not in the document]
ArekBalysNordic: Done.

scripts/west/west-commands.yml
ArekBalysNordic: Changed to zap-append :)

scripts/west/zap_gui.py
```diff
@@ -31,33 +33,55 @@ def do_add_parser(self, parser_adder):
...
+        app_templates_path = DEFAULT_APP_TEMPLATES_PATH
+
+        print(f"Using ZAP file: {zap_file_path}")
+        print(f"Using ZCL file: {zcl_json_path}")
+        print(f"Using app templates: {app_templates_path}")
 
         zap_installer = ZapInstaller(args.matter_path)
         zap_installer.update_zap_if_needed()
-        zap_cache_path = zap_installer.get_install_path() / ".zap"
+        zap_cache_path = zap_installer.get_cache_path()
```
ArekBalysNordic, line 68: I will add here temporary cache directory instead of using the global one inside the installation dir.

scripts/west/zap_append.py
```diff
@@ -0,0 +1,98 @@
...
+        if not Path(cluster).exists():
+            raise RuntimeError(f"No such cluster file: {cluster}")
+
+        file = Path(cluster).name
+        if not output:
+            output = DEFAULT_ZCL_PATH
+
+        relative_path = Path(cluster).absolute().parent.relative_to(output.parent, walk_up=True)
+
+        # We need to add to things:
```
kkasperczyk-no, line 58:
```suggestion
        # We need to add two things:
```

scripts/west/zap_append.py
```diff
@@ -0,0 +1,98 @@
...
+                                         formatter_class=argparse.RawDescriptionHelpFormatter,
+                                         description=self.description)
+        parser.add_argument("-b", "--base", type=Path, default=DEFAULT_ZCL_PATH,
+                            help=f"An absolute path to the base zcl.json file. If not provided the path will be set to {DEFAULT_ZCL_PATH}.")
+        parser.add_argument("-m", "--matter", type=Path, default=DEFAULT_MATTER_PATH,
+                            help=f"An absolute path to the Matter directory. If not set the path with be set to the {DEFAULT_MATTER_PATH}")
+        parser.add_argument("-o", "--output", type=Path, default=DEFAULT_ZCL_PATH,
+                            help=f"Output path to store the generated zcl.json file. If not provided the path will be set to the base zcl.json file ({DEFAULT_ZCL_PATH}).")
+        parser.add_argument("new_clusters", nargs='+',
+                            help="Paths to the XML files that contains the custom cluster definitions")
```
kkasperczyk-no, line 94:
```suggestion
                            help="Paths to the XML files that contain the custom cluster definitions")
```

scripts/west/zap_gui.py
```diff
@@ -31,33 +33,52 @@ def do_add_parser(self, parser_adder):
                                          description=self.description)
         parser.add_argument('-z', '--zap-file', type=existing_file_path,
                             help='Path to data model configuration file (*.zap)')
-        parser.add_argument('-j', '--zcl-json', type=existing_file_path,
-                            help='Path to data model definition file (zcl.json)')
+        parser.add_argument('-j', '--zcl-json', type=str,
+                            help='Path to data model definition file (zcl.json). If new clusters are added using --clusters, the new zcl.json file will be be created and used.')
```
kkasperczyk-no, line 37: It will be created if there isn't one, right? Otherwise the existing zcl.json will be updated.
```suggestion
                            help='Path to data model definition file (zcl.json). If new clusters are added using --clusters, the new zcl.json file will be created and used.')
```

scripts/west/zap_gui.py
```diff
@@ -31,33 +33,52 @@ def do_add_parser(self, parser_adder):
...
         cmd += [zap_file_path] if zap_file_path else []
         cmd += ["--zcl", zcl_json_path]
         cmd += ["--gen", app_templates_path]
-        cmd += ["--stateDirectory", zap_cache_path]
-
+        cmd += ["--stateDirectory", zap_installer.get_cache_path()]
         self.check_call([str(x) for x in cmd])
+
+        # clear cache after the usage
+        zap_installer.clear_cache()
```
kkasperczyk-no, line 84: Shouldn't we check `if args.clear` before that?

ArekBalysNordic: I forgot to remove args.clear :) Now cache is always temporary.

scripts/west/zap_common.py
```diff
@@ -3,14 +3,15 @@
...
-import os
 import platform
 import re
 import shutil
 import stat
 import subprocess
 import tempfile
 import wget
+import json
+import tempfile
```
Damian-Nordic, line 14: Duplicate

scripts/west/zap_common.py
```diff
@@ -77,13 +81,43 @@ def existing_dir_path(arg: str) -> Path:
...
+
+
 class ZapInstaller:
     INSTALL_DIR = Path('.zap-install')
     ZAP_URL_PATTERN = 'https://github.com/project-chip/zap/releases/download/v%04d.%02d.%02d-nightly/%s.zip'
 
     def __init__(self, matter_path: Path):
         self.matter_path = matter_path
         self.install_path = matter_path / ZapInstaller.INSTALL_DIR
+        self.cache_path = tempfile.mkdtemp()
```
Damian-Nordic, line 120: You can use ZAP's `--tempState` instead of `--stateDirectory` with a manually created temp dir.

scripts/west/zap_gui.py
```diff
@@ -31,33 +33,55 @@ def do_add_parser(self, parser_adder):
...
 
         zap_installer = ZapInstaller(args.matter_path)
         zap_installer.update_zap_if_needed()
-        zap_cache_path = zap_installer.get_install_path() / ".zap"
+        zap_cache_path = zap_installer.get_cache_path()
+
+        # The zcl.json path in the .zap file must be the same as the one provided by the user
+        # If not, update the .zap file with the new relative path to the zcl.json file.
+        # After that we must clear the ZAP cache.
+        if not check_zcl_in_zap(zap_file_path, zcl_json_path):
```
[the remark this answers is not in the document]
ArekBalysNordic: I was also surprised, but it doesn't work this way... It takes zcl file that is provided in the zap file if the zap file is provided as an argument. If we run the ZAP tool without providing the zap file directly, it uses it as you wrote.

scripts/west/zap_common.py
ArekBalysNordic: Hmm, oh nice.

scripts/west/zap_append.py
```diff
@@ -0,0 +1,98 @@
...
+            'A tool for adding a custom cluster to the ZCL Matter Data Model file according to the base ZCL file and custom clusters definitions.')
+
+    def do_add_parser(self, parser_adder) -> argparse.ArgumentParser:
+        parser = parser_adder.add_parser(self.name,
+                                         help=self.help,
+                                         formatter_class=argparse.RawDescriptionHelpFormatter,
+                                         description=self.description)
+        parser.add_argument("-b", "--base", type=Path, default=DEFAULT_ZCL_PATH,
+                            help=f"An absolute path to the base zcl.json file. If not provided the path will be set to {DEFAULT_ZCL_PATH}.")
+        parser.add_argument("-m", "--matter", type=Path, default=DEFAULT_MATTER_PATH,
```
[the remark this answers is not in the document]
ArekBalysNordic: Converted all occurrences to respect a different matter SDK path than the default one. :)

scripts/west/zap_common.py
ArekBalysNordic: Done.

scripts/west/zap_gui.py
```diff
@@ -31,33 +34,55 @@ def do_add_parser(self, parser_adder):
...
+        parser.add_argument('-c', '--cache', type=Path,
+                            help='Path to the custom cache directory. If not provided a temporary directory will be used and cleared after the usage.')
         return parser
 
     def do_run(self, args, unknown_args):
-        if args.zap_file:
-            zap_file_path = args.zap_file
-        else:
-            zap_file_path = find_zap()
+        DEFAULT_ZCL_PATH = args.matter_path.joinpath(DEFAULT_ZCL_JSON_RELATIVE_PATH)
```
adigie, line 48: Uppercase names are usually used for constants, it may be misleading.

adigie (approved):

maciejbaczmanski (approved):

PR #551, opened 2025-02-11
Zephyr native tests

adigie:
Run Zephyr native tests in CI.

Following line configures workflow to use NCS from specified PR:
NCS PR: nrfconnect/sdk-nrf/pull/20287


LuDuda (approved):
LGTM :+1:

Btw. Not sure.. but could we squash noups to single commit for maintenance reasons?

adigie:
> Btw. Not sure.. but could we squash noups to single commit for maintenance reasons?

Sure.

maciejbaczmanski (approved):

PR #566, opened 2025-03-03
Remove unused submodules

adigie:
This reduces size of the repo to ~1.8GB.


.gitmodules
```diff
@@ -27,321 +18,22 @@
 	path = third_party/openthread/repo
```
ArekBalysNordic, line 18: Do we even need openthread repo there? We actually use openthread from Zephyr.

.gitmodules
```diff
@@ -27,321 +18,22 @@
...
-	platforms = silabs
-[submodule "third_party/silabs/wifi_sdk"]
-	path = third_party/silabs/wifi_sdk
-	url = https://github.com/SiliconLabs/wiseconnect.git
-	branch = v3.3.3
-	platforms = silabs
 [submodule "editline"]
 	path = third_party/editline/repo
 	url = https://github.com/troglobit/editline.git
 [submodule "third_party/boringssl/repo/src"]
```
ArekBalysNordic, line 27: Is this used for some tests? I though that we use PSA or Mbedtls only.

adigie: We use `boringssl` to build tools.

.gitmodules
adigie: Build system is complaining when it's not there. I will check if I can get rid of it easily.

.gitmodules
adigie: I think `mbedtls` does not support some feature that we need in one of the tools, so we need `openssl` or `boringssl`.

ArekBalysNordic: ok, so leave it as it is :)

ArekBalysNordic (approved):

PR #572, opened 2025-03-12
[nrf fromtree] Add factory reset event

adigie:
Add `kFactoryReset` event and post it at the beginning of `ScheduleFactoryReset`.

### DIFF
--- a/src/app/server/Server.cpp
+++ b/src/app/server/Server.cpp
@@ -586,8 +586,22 @@ void Server::GenerateShutDownEvent()
     PlatformMgr().ScheduleWork([](intptr_t) { PlatformMgr().HandleServerShuttingDown(); });
 }
 
+void Server::PostFactoryResetEvent()
+{
+    DeviceLayer::ChipDeviceEvent event;
+    event.Type = DeviceLayer::DeviceEventType::kFactoryReset;
+
+    CHIP_ERROR error = DeviceLayer::PlatformMgr().PostEvent(&event);
+    if (error != CHIP_NO_ERROR)
+    {
+        ChipLogError(AppServer, "Posting kFactoryReset event failed with %" CHIP_ERROR_FORMAT, error.Format());
+    }
+}
+
 void Server::ScheduleFactoryReset()
 {
+    PostFactoryResetEvent();
+
     PlatformMgr().ScheduleWork([](intptr_t) {
         // Delete all fabrics and emit Leave event.
         GetInstance().GetFabricTable().DeleteAllFabrics();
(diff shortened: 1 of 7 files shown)


src/app/server/Server.cpp
```diff
@@ -586,8 +586,22 @@ void Server::GenerateShutDownEvent()
...
 
+void Server::PostFactoryResetEvent()
+{
+    DeviceLayer::ChipDeviceEvent event;
+    event.Type = DeviceLayer::DeviceEventType::kFactoryReset;
+
+    CHIP_ERROR error = DeviceLayer::PlatformMgr().PostEvent(&event);
+    if (error != CHIP_NO_ERROR)
+    {
+        ChipLogError(Discovery, "Posting kFactoryReset event failed with %" CHIP_ERROR_FORMAT, error.Format());
```
ArekBalysNordic, line 597: Discovery?

ArekBalysNordic (commented):
Overall looks good. What about the previous workaround from here: https://github.com/nrfconnect/sdk-connectedhomeip/pull/569. Once we merge this PR, we don't need to have that one, right? So if you're going to merge this one, maybe it doesn't make sense to merge that one. Otherwise, we will have two different approaches for handling factory resets in the application. Or if you want to merge that one anyway, let's open this PR on upstream first and see what the community thinks about this instead of making it [toup].

src/app/server/Server.cpp
adigie: I copied posting event from another file, will fix this.

adigie:
> Overall looks good. What about the previous workaround from here: #569. Once we merge this PR, we don't need to have that one, right? So if you're going to merge this one, maybe it doesn't make sense to merge that one. Otherwise, we will have two different approaches for handling factory resets in the application. Or if you want to merge that one anyway, let's open this PR on upstream first and see what the community thinks about this instead of making it [toup].

This is alternative approach to the #569. We can merge this instead. I'm not sure if this would be accepted on upstream. It will probably need to be a noup.

ArekBalysNordic:
> > Overall looks good. What about the previous workaround from here: #569. Once we merge this PR, we don't need to have that one, right? So if you're going to merge this one, maybe it doesn't make sense to merge that one. Otherwise, we will have two different approaches for handling factory resets in the application. Or if you want to merge that one anyway, let's open this PR on upstream first and see what the community thinks about this instead of making it [toup].
> 
> This is alternative approach to the #569. We can merge this instead. I'm not sure if this would be accepted on upstream. It will probably need to be a noup.

Ok, so this approach seems better to me than the previous one. I think it doesn't seem so bad, and it is clear, maybe we should try to put it in upstream. :)

PR #573, opened 2025-03-14
[nrf fromtree] OpenThread: Clear SRP host and services during RevertConfiguration

adigie:
* Postpone `RevertConfiguration` until OpenThread is initialized.
* Clear SRP host and services before reverting dataset.
* Abort revert if creating new backup.

### DIFF
--- a/src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
+++ b/src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
@@ -51,8 +51,13 @@ CHIP_ERROR GenericThreadDriver::Init(Internal::BaseDriver::NetworkStatusChangeCa
 
     // If the network configuration backup exists, it means that the device has been rebooted with
     // the fail-safe armed. Since OpenThread persists all operational dataset changes, the backup
-    // must be restored on the boot. If there's no backup, the below function is a no-op.
-    RevertConfiguration();
+    // must be restored.
+    if (BackupExists())
+    {
+        // Set flag and postpone revert until OpenThread is initialized,
+        // as we need to clear SRP host and services before restoring the backup.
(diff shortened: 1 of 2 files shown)


src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
```diff
@@ -51,8 +51,13 @@ CHIP_ERROR GenericThreadDriver::Init(Internal::BaseDriver::NetworkStatusChangeCa
 
     // If the network configuration backup exists, it means that the device has been rebooted with
     // the fail-safe armed. Since OpenThread persists all operational dataset changes, the backup
-    // must be restored on the boot. If there's no backup, the below function is a no-op.
-    RevertConfiguration();
+    // must be restored.
+    if (BackupExists())
```
Damian-Nordic, line 55: I don't think this check is necessary. `RevertConfiguration` takes care on its own whether the backup exists or not and `kServerReady` is generated only once per boot, so `mRevertOnStartup` seems redundant, right?

Damian-Nordic (commented):
I think now that you delay `RevertConfiguration`, it needs to be considered what happens if `BackupConfiguration` is called before `kServerReady` is processed (which may happen if SRP response gets delayed or is lost and we receive a Matter command to modify Thread network). Then we should probably do the revert without waiting for the SRP response.

src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
adigie: Discussed offline - we will keep `mRevertOnStartup` to mark that during initialization backup was present. Additionally, we have to clear it when creating a new backup.

PR #578, opened 2025-03-20
Fix examples configuration

adigie:


examples/platform/nrfconnect/util/OTAUtil.cpp
```diff
@@ -93,9 +93,3 @@ void OtaConfirmNewImage()
 }
 
 #endif
-
-ExternalFlashManager & GetFlashHandler()
```
ArekBalysNordic, line 97: If so, let's remove it from OTAUtil.h as well.

ArekBalysNordic (approved):

PR #579, opened 2025-03-20
Fix zap-gui and add zap-install

adigie:


scripts/west/zap_gui.py
```diff
@@ -50,6 +50,10 @@ def do_run(self, args, unknown_args):
         zap_file_path = args.zap_file or find_zap()
         zcl_json_path = Path(args.zcl_json).absolute() if args.zcl_json else default_zcl_path
 
+        if zap_file_path is None:
```
Damian-Nordic, line 53: What exactly was broken? Initially this was deliberately supported to run zap-gui without the file to just play around with the gui :)

adigie: When it's run from west workspace root then it fails with:
```
$ west zap-gui
Using ZAP file: None
Using ZCL file: /home/adi/ncs/main/modules/lib/matter/src/app/zap-templates/zcl/zcl.json
Using app templates: /home/adi/ncs/main/modules/lib/matter/src/app/zap-templates/app-templates.json
[1172501:0320/123810.900453:FATAL:setuid_sandbox_host.cc(158)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /home/adi/ncs/main/modules/lib/matter/.zap-install/chrome-sandbox is owned by root and has mode 4755.
ZAP installation directory: /home/adi/ncs/main/modules/lib/matter/.zap-install
Installing ZAP 2024.8.14
Downloading https://github.com/project-chip/zap/releases/download/v2024.08.14-nightly/zap-linux-x64.zip...
100% [......................................................................] 209154442 / 209154442
Unzipping ZAP package to /home/adi/ncs/main/modules/lib/matter/.zap-install...
Traceback (most recent call last):
  File "/home/adi/ncs/toolchains/aedb4c0245/usr/local/bin/west", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/adi/ncs/toolchains/aedb4c0245/usr/local/lib/python3.12/site-packages/west/app/main.py", line 1085, in main
    app.run(argv or sys.argv[1:])
  File "/home/adi/ncs/toolchains/aedb4c0245/usr/local/lib/python3.12/site-packages/west/app/main.py", line 244, in run
    self.run_command(argv, early_args)
  File "/home/adi/ncs/toolchains/aedb4c0245/usr/local/lib/python3.12/site-packages/west/app/main.py", line 505, in run_command
    self.run_extension(args.command, argv)
  File "/home/adi/ncs/toolchains/aedb4c0245/usr/local/lib/python3.12/site-packages/west/app/main.py", line 654, in run_extension
    self.cmd.run(args, unknown, self.topdir, manifest=self.manifest,
  File "/home/adi/ncs/toolchains/aedb4c0245/usr/local/lib/python3.12/site-packages/west/commands.py", line 194, in run
    self.do_run(args, unknown)
  File "/home/adi/ncs/main/modules/lib/matter/scripts/west/zap_gui.py", line 76, in do_run
    was_updated = update_zcl_in_zap(zap_file_path, zcl_json_path, app_templates_path)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adi/ncs/main/modules/lib/matter/scripts/west/zap_common.py", line 92, in update_zcl_in_zap
    with open(zap_file, 'r+') as file:
         ^^^^^^^^^^^^^^^^^^^^
TypeError: expected str, bytes or os.PathLike object, not NoneType
```
Changed the fix to skip calling `update_zcl_in_zap` if ZAP file was not found.

scripts/west/zap_gui.py
```diff
@@ -70,12 +70,13 @@ def do_run(self, args, unknown_args):
         zap_installer = ZapInstaller(args.matter_path)
         zap_installer.update_zap_if_needed()
 
-        # The zcl.json path in the .zap file must be the same as the one provided by the user
-        # If not, update the .zap file with the new relative path to the zcl.json file.
-        # After that we must clear the ZAP cache.
-        was_updated = update_zcl_in_zap(zap_file_path, zcl_json_path, app_templates_path)
-        if args.cache and was_updated:
-            log.wrn("ZCL file path in the ZAP file has been updated. The ZAP cache must be cleared to use it.")
+        if zap_file_path is not None:
```
ArekBalysNordic, line 73: Why? If someone provides zap_file_path manually, the zcl path inside it must be updated as well. Otherwise, the Zap tool gets the wrong paths, and, for example, new clusters are not taken into account. Moreover, I don't see an advantage of using this script outside the sample's directory. Our documentation is about this approach.

ArekBalysNordic (changes requested):

scripts/west/zap_gui.py
adigie: Because currently it just fails and print traceback. You can still open specific file when provided path with `-z` argument.

I added this to just ignore the ZAP file if it's not found and open the GUI. Please see Damian's comment.

Damian-Nordic (approved):

scripts/west/zap_gui.py
ArekBalysNordic: The problem is if someone wants to open zap GUI externally, not from the sample's directory and zap file doesn't exist, then adding zcl path won't work. So, it is ok, but we should add a note to the guide for that because it may lead to problems. Alternatively, what seems better, we should prevent using the `--zcl` argument if we call without providing a zap file.

Damian-Nordic: @adigie I think you can restore the previous version and just not support this use case anymore :) I might have missed that this command got more complex and is now doing more than a single thing, so maybe let's not complicate it further :). Sorry for the confusion!

adigie: Restored previous version, `west zap-gui` will print `ERROR: ZAP file not found!` and exit if it finds no ZAP file.

ArekBalysNordic (approved):

maciejbaczmanski (approved):

PR #581, opened 2025-03-24
[nrf noup] Optimized configuration for LITs

kkasperczyk-no:
The LIT configuration in Matter samples can be slightly changed to achieve lower power consumption.


ArekBalysNordic (approved):

kkasperczyk-no:
After having a look at the failing workflow on native posix, it seems it's not related to this PR. There is a problem with unit test and we are going to work on fixing it separately. I'm merging with a red workflow, because CI job in the sdk-nrf passed, so the risk is minimal.

PR #583, opened 2025-03-29
[nrfconnect] Reconfigure memory pool sizes

rado17:
With introduction of dedicated memory pools for nRF70 Wi-Fi driver, redistribute current heap size among the
driver heap and kernel heap.

> Add following line to test this PR against different NCS revision:
> NCS PR: nrfconnect/sdk-nrf/pull/<PR ID>

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig.defaults
+++ b/config/nrfconnect/chip-module/Kconfig.defaults
@@ -53,8 +53,17 @@ config SYSTEM_WORKQUEUE_STACK_SIZE
     default 2560 if CHIP_WIFI
     default 2048 if CHIP_ICD_LIT_SUPPORT
 
+if CHIP_WIFI
+
 config HEAP_MEM_POOL_SIZE
-    default 110720 if CHIP_WIFI
+    default 48336 if BUILD_WITH_TFM
+    default 40144
+
+config NRF_WIFI_DATA_HEAP_SIZE
+    default 42384 if BUILD_WITH_TFM
+    default 50576
+
+endif # CHIP_WIFI
 
 config HEAP_MEM_POOL_IGNORE_MIN
     default y if CHIP_WIFI


config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -53,9 +53,18 @@ config SYSTEM_WORKQUEUE_STACK_SIZE
...
+    default 48336 if BUILD_WITH_TFM
+    default 40144
+
+config NRF_WIFI_DATA_HEAP_SIZE
+    default 50576 if BUILD_WITH_TFM
+    default 42384
 
+config NRF_WIFI_CTRL_HEAP_SIZE
+    default 20000
+endif # CHIP_WIFI
```
krish2718, line 67: better to leave this default, no? It shouldn't be controlled by samples?

rado17: Agreed. Done.

krish2718 (approved):

config/nrfconnect/chip-module/Kconfig.defaults
```diff
@@ -53,8 +53,17 @@ config SYSTEM_WORKQUEUE_STACK_SIZE
...
 
+if CHIP_WIFI
+
 config HEAP_MEM_POOL_SIZE
-    default 110720 if CHIP_WIFI
+    default 48336 if BUILD_WITH_TFM
+    default 40144
+
+config NRF_WIFI_DATA_HEAP_SIZE
+    default 50576 if BUILD_WITH_TFM
```
kkasperczyk-no, line 63: Where do these values come from? The description says "redistribute", but the given sizes do not sum up to 110720. Are you sure it's safe to decrease the pool size?

LuDuda (approved):
If the values are confirmed then, it's great to see the optimization!

I would run the nightly tests on it (we might do it also after the merge?) - @doublemis1

config/nrfconnect/chip-module/Kconfig.defaults
kkasperczyk-no: @rado17 ?

rado17: Need to correct the `NRF_WIFI_DATA_HEAP_SIZE`. The values are interchanged. With `BUILD_WITH_TFM` where `MBEDTLS_HEAP` will be disabled, we need to reduce the `NRF_WIFI_DATA_HEAP_SIZE` by 8192 bytes (to 42384).  Pushed the changes now.
`NRF_WIFI_CTRL_HEAP_SIZE` is set to 20000 by default. The total comes to 110720.

rado17: We have done some basic performance tests with `NRF_WIFI_DATA_HEAP_SIZE` set to 40000 and `NRF_WIFI_CTRL_HEAP_SIZE` set to 20000 and didn't see any issues. Reserving 40144 bytes for K_HEAP is required for supplicant and socketpair subsystems.

kkasperczyk-no (approved):

PR #599, opened 2025-04-22
[nrf noup] Add --keep-previous argument to zap-generate

adigie:
* Remove previous ZAP generated files.
* Add `-k`/`--keep-previous` argument to skip removing old files.


scripts/west/zap_generate.py
```diff
@@ -81,3 +85,14 @@ def do_run(self, args, unknown_args):
...
 
         log.inf(f"Done. Files generated in {output_path}")
+
+    def clear_generated_files(self, path: Path):
+        for file in path.iterdir():
+            if file.is_file():
+                with open(file, 'r') as f:
+                    for line in f.readlines():
+                        if "// THIS FILE IS GENERATED BY ZAP" in line:
+                            print(f"Removing {file}")
```
ArekBalysNordic, line 95:
```suggestion
                            log.dbg(f"Removing {file}")
```

To support west logging levels.

ArekBalysNordic (approved):

LuDuda (approved):
Looks good :+1:

Just a thought.. I guess this is more safe - but would it make sense to remove previously generated files by default?

adigie:
> Looks good 👍
> 
> Just a thought.. I guess this is more safe - but would it make sense to remove previously generated files by default?

Sure, I guess that if we filter the files by `// THIS FILE IS GENERATED BY ZAP` then it should be safe to remove them by default. @ArekBalysNordic what do you think?

ArekBalysNordic:
> > Looks good 👍
> > Just a thought.. I guess this is more safe - but would it make sense to remove previously generated files by default?
> 
> Sure, I guess that if we filter the files by `// THIS FILE IS GENERATED BY ZAP` then it should be safe to remove them by default. @ArekBalysNordic what do you think?

I have no strong opinion on that, both ways are ok, and that's safe indeed. If we decide to make it default, thenwe should write a release note entry, because we're changing the default behaviour of the script that people are using right now.

scripts/west/zap_generate.py
```diff
@@ -37,6 +38,7 @@ def do_add_parser(self, parser_adder):
         parser.add_argument('-m', '--matter-path', type=existing_dir_path,
                             default=DEFAULT_MATTER_PATH, help='Path to Matter SDK')
         parser.add_argument('-f', '--full', action='store_true', help='Generate full data model files')
+        parser.add_argument('-k', '--keep-previous', action='store_true', help='Keep previously generated files')
```
ArekBalysNordic, line 41: `'store_false'`?

PR #600, opened 2025-04-24
[nrf noup] Use ZAP CLI for version check

kendallgoto:
Both `zap --version` and `zap-cli --version` do the same thing more or less, but `zap` is heavier to run since it starts Electron just to do `--version`. Since the user may only care about `zap-cli` commands (e.g. `west zap-generate`), requiring that all `west zap-*` commands call `zap` just to get a version check increases the amount of resources and dependencies required.

Instead, just use `zap-cli` to get the version.

### DIFF
--- a/scripts/west/zap_common.py
+++ b/scripts/west/zap_common.py
@@ -189,13 +189,13 @@ class ZapInstaller:
         """
         Returns ZAP package current version as a tuple of integers.
 
-        Parses the output of `zap --version` to determine the current ZAP
+        Parses the output of `zap-cli --version` to determine the current ZAP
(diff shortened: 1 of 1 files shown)


Damian-Nordic (approved):
Thanks for the contribution! I agree it makes sense.

kendallgoto:
Do I need to open a manifest PR in `sdk-nrf` as per the failing check or are you able to take it from here? Thanks!

adigie:
Hi @kendallgoto, you have to sign Contributor License Agreement before we can merge this. Please force push to this PR (and #601) to trigger the CLAassistant.

ArekBalysNordic (approved):

PR #601, opened 2025-04-24
[nrf noup] Use proper ZAP package for Linux/Windows with ARM CPUs

kendallgoto:
Currently x86 ZAP is always downloaded on Linux & Windows. This change looks at the architecture of the system to download either x86_64 or arm64 binaries as provided by ZAP releases

This is the Linux / Windows equivalent to #574 which similarly downloads ARM64 binaries for Darwin systems on Apple Silicon.

### DIFF
--- a/scripts/west/zap_common.py
+++ b/scripts/west/zap_common.py
@@ -132,12 +132,18 @@ class ZapInstaller:
             f.close()
 
         if self.current_os == 'Linux':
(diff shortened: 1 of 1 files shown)


kkasperczyk-no (approved):
It looks good to me :+1: Thank you for the contribution!

ArekBalysNordic (approved):

PR #608, opened 2025-05-09
[nrf toup] [Groups] Do not add group if there is no listener

ArekBalysNordic:
Check if the listener has been assigned to the multicast group; if not, remove the group and return `ResourceExhausted` status in the AddGroupResponse message.


LuDuda:
Close until enhancements are done in Matter SDK.

PR #609, opened 2025-05-12
[nrf toup] [Crypto] Use single-part API for PSA AEAD functions.

ArekBalysNordic:
Use single-part instead of multi-part API for PSA AEAD functions for AES_CCM encrypt and decrypt methods. This reduces amount of crypto interactions and keeps the same functionality.

manifest-pr-skip


LuDuda:
Let's close for now. We can re-open if we decide to go with dynamic memory variant instead.

PR #613, opened 2025-05-17
[nrf toup] Disable BLE SMP buffer scaling on nRF54L10

LuDuda:
Disable automatic increase of the Bluetooth SMP MTU and RX buffers on the nRF54L10 SoC.

### DIFF
--- a/config/nrfconnect/chip-module/Kconfig.features
+++ b/config/nrfconnect/chip-module/Kconfig.features
@@ -184,10 +184,10 @@ config MCUMGR_SMP_COMMAND_STATUS_HOOKS
 
 # Increase BT MTU and RX buffer for big size DFU messages
 config BT_L2CAP_TX_MTU
-	default 498
+	default 498 if !SOC_NRF54L10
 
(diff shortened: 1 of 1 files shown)


kkasperczyk-no (commented):
Why don't we need it only for 54L10? Or is there some issue with RAM usage and we want to decrease it?

LuDuda:
@kkasperczyk-no would it work like this, or you prefer to align all the board files to force smaller buffers? Note that this configuration is enabled only when CHIP BLE SMP is enabled.

kkasperczyk-no:
> @kkasperczyk-no would it work like this, or you prefer to align all the board files to force smaller buffers? Note that this configuration is enabled only when CHIP BLE SMP is enabled.

Ok, so using board files could not be the perfect choice. I'm ok to keep it as is, but please add a comment that it was decreased because of them RAM usage. Otherwise someone may wonder why this specific board requires different configuration for SMP.

adigie (approved):

kkasperczyk-no (approved):

PR #615, opened 2025-05-19
[nrf noup] Build chip-tool with logs

adigie:
Add `enable_im_pretty_print=true` to enable detailed debug logs.

manifest-pr-skip


LuDuda (approved):

Damian-Nordic:
I thought `is_debug` not only enables debug logs but also enables C asserts and disable some optimizations. Isn't there more specific flag for just enabling logs that you need?

doublemis1:
I remember some config like PRETTY_PRINT or something similar

Damian-Nordic (approved):

PR #621, opened 2025-06-25
Enabled Matter operation with Openthread and without Zephyr L2

kkasperczyk-no:
Cherry-picked enabler commit from Matter upstream and added few required fixes.

### Testing
Automated tests in CI


src/platform/Zephyr/ThreadStackManagerImpl.cpp
```diff
@@ -52,18 +52,18 @@ CHIP_ERROR ThreadStackManagerImpl::_InitThreadStack()
 
 void ThreadStackManagerImpl::_LockThreadStack()
 {
-    openthread_api_mutex_lock(openthread_get_default_context());
+    openthread_mutex_lock();
```
ArekBalysNordic, line 55: It can cause problems in the upstream, because it is valid only from a specific Zephyr version.

adigie: You can include `zephyr/version.h` and then use `KERNEL_VERSION_MAJOR` and `KERNEL_VERSION_MINOR`. I think this API is available in `4.1`.

adigie (approved):

PR #632, opened 2025-07-30
[nrf fromlist][dnssd] Fixed buffer sizes for dns names

kkasperczyk-no:
#### Summary
The DNS browse for vendor id subtype does not work, as full name does not fit in the provided buffer.

By the way of investigating this issue couple of other issues were found:

* Started to use full service name including subtypes for dns browse results handling in Thread implementation
* Fixed max commissionable and commissioner service name sizes that didn't include . character
* Fixed kDnssdFullTypeAndProtocolMaxSize that did not include _sub phrase

#### Testing
Tested manually using `dns browse` command from the dns shell for the _V and _L subtypes.


ArekBalysNordic (approved):

kkasperczyk-no:
> Btw. Shouldn't this be [nrf fromlist] as you have contributed it as well.
yes, I've just pushed PR to upstream afterwards. Once it will be approved there I will change this one to fromtree/fromlist

PR #640, opened 2025-08-20
[nrf noup] Workaround for the KMU usage with the new NOC approach

ArekBalysNordic:
Now, we copy the NOC key from volatile to persistent space during the commissioning OP keypair.

Currently, psa_copy_key does not work for KMU, so we must export this key and import it again as a workaround.

This PR is a workaround and can be fully reverted once the NCSDK-34997 issue is resolved.

#### Testing

In NCS.

manifest-pr-skip


doublemis1:
@ArekBalysNordic should we address it or it is currently blocked?

ArekBalysNordic:
> @ArekBalysNordic should we address it or it is currently blocked?

Still waiting for the review :)

PR #644, opened 2025-09-30
[nrf toup] [nrfconnect] Remove dependency to DK in Matter configuration

ArekBalysNordic:
In Matter configuration there must not be anything related to DK. All DK-related Kconfigs must be defined in the application.

#### Testing

NRF

manifest-pr-skip


config/nrfconnect/chip-module/Kconfig.features
```diff
@@ -21,7 +21,7 @@ if CHIP
 
 config CHIP_QSPI_NOR
 	bool "Enable QSPI NOR feature set"
-	imply NORDIC_QSPI_NOR
+	default y if NORDIC_QSPI_NOR
```
kkasperczyk-no, line 24: I think this symbol becomes pointless now. It was created to enable required nrf/zephyr symbols from Matter. If you want to set it in a reversed order, just use `NORDIC_QSPI_NOR` in .cpp code and remove this symbol at all.

config/nrfconnect/chip-module/Kconfig.features
```diff
@@ -35,9 +35,7 @@ endif # CHIP_QSPI_NOR
 
 config CHIP_SPI_NOR
 	bool "Enable SPI NOR feature set"
-	imply SPI
-	imply SPI_NOR
-	imply MULTITHREADING
+	default y if SPI_NOR
```
kkasperczyk-no, line 38: Ditto, I don't think we need this symbol if it's going to be set by nrf symbol.

kkasperczyk-no: Btw who is going to set `SPI_NOR` now? I don't see it in Matter samples

config/nrfconnect/chip-module/Kconfig.features
```diff
@@ -130,8 +128,6 @@ endif
 
 config CHIP_DFU_OVER_BT_SMP
 	bool "Enable DFU over Bluetooth LE SMP feature set"
-	imply CHIP_QSPI_NOR if BOARD_NRF5340DK_NRF5340_CPUAPP || BOARD_NRF52840DK_NRF52840
```
kkasperczyk-no, line 133: The same question, who is going to take care of enabling SPI_NOR/QSPI_NOR? Our documentation says it's enough to enable this config to support DFU, but now it is not going to be true.

config/nrfconnect/chip-module/Kconfig.features
ArekBalysNordic: Driver itself according to DTS

kkasperczyk-no: Cool, it must be something "new", because few years back it was not working like this.

PR #718, opened 2026-04-28
Matter changes for Mbed TLS update to v4.1.0

tomi-font:
manifest-pr-skip

#### Testing

NCS PR: nrfconnect/sdk-nrf/pull/28496


LuDuda (approved):
LGTM :+1:

@kkasperczyk-no can you take a look from commits perspective? There are few noups and toups. At some point we would need to upstream the commits -> though upstream might need some #ifdefs for mbedTLS version as paths changed..

tomi-font:
Thanks for having a look @LuDuda!
@Vge0rge is still working on the remaining failures for OT/Matter.
I think indeed his commits should be marked as toup, but we will let you manage.

And yes, definitely take a look at the changes we made (in the `nrfxlib` PR too) and please make sure that things run properly too. :slightly_smiling_face:

kkasperczyk-no:
> LGTM 👍
> 
> @kkasperczyk-no can you take a look from commits perspective? There are few noups and toups. At some point we would need to upstream the commits -> though upstream might need some #ifdefs for mbedTLS version as paths changed..

Unless there is a very important reason for keeping it as a nordic patch I would suggest making all commits as nrf toups, so we could contribute it and would not need to maintain the no up in the future. 

And yes, if changing include paths is required probably it would be a good idea to use ifdef with mbedtls version, otherwise it will fail for the other platforms that have not switched to newer version yet. Alternative solution is using ifdef for CHIP_PLATFORM_NRFCONNECT, but this is a last resort solution in the common files.

LuDuda (approved):
LGTM :+1:

Changes seems fine and reasonable. Thanks.

PR #719, opened 2026-04-29
[nrf noup] [west zap] Refinements and fixes to some edge cases

markaj-nordic:
[nrf noup] [west zap-gui] Refine error handling

It turns out that the subprocess may return no output
if the Chromium cannot use sandbox due to the missing
permissions. Added a workaround that raises an exception
if the command execution output is empty, so that it
can be handled with the fix_sandbox_permissions() call.

[nrf noup] [west zap-sync] Handle overwriting of the default zcl file

Fixed the handling of the case when the default
cluster definition file (zcl.json) is used as both
input and output to the command. Without the fix,
the script removes the zcl.json and exits, because
the removed file cannot be copied.

[nrf noup] [west zap-append] Add fallback to the attribute name deduction

Some XML cluster definitions do not contain the dedicated
'name' entry in the attribute field, instead they just
add name of the cluster as a text to the attribute
definition. Added a fallback logic, so that such cases
are handled to avoid ending up with null attributes in
zcl.json file. This mimics the ZAP loader behavior.

#### Testing

Tested manually when adding a custom cluster.

### DIFF
--- a/scripts/west/zap_append.py
+++ b/scripts/west/zap_append.py
@@ -12,6 +12,15 @@ from west.commands import WestCommand
 from zap_common import DEFAULT_MATTER_PATH, DEFAULT_MATTER_TYPES_RELATIVE_PATH, DEFAULT_ZCL_JSON_RELATIVE_PATH
 
 
+def get_attribute_name(attribute: ET.Element) -> str | None:
+    """If attribute has a name, return it, otherwise fallback to the text of the attribute entry."""
+    from_attr = attribute.get('name')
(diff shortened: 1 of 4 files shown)


kkasperczyk-no (approved):
LGTM, worth to address the copilot comments and checking if it works on desk.

markaj-nordic:
> LGTM, worth to address the copilot comments and checking if it works on desk.

The valid Copilot comments have been addressed.

ArekBalysNordic (approved):

PR #722, opened 2026-05-25
[nrf toup][nrfconnect] enable Secure Storage instead of Trusted Storage

tomi-font:
Trusted Storage is deprecated in favor of Secure Storage, so make CONFIG_CHIP_WIFI_CRYPTO_BACKEND_PSA default to Secure Storage.

Trusted Storage remains a supported option for the time being so keep enabling the same Kconfig options when Trusted Storage is used. When Secure Storage is used, nothing more needs to be enabled as the build system handles enabling what is needed automatically.

#### Testing

NCS PR: nrfconnect/sdk-nrf/pull/28996

manifest-pr-skip


tomi-font:
@nrfconnect/ncs-matter please review

PR #723, opened 2026-05-27
BLE advertising improvements

adigie:
#### Summary

This PR adds multi BT indentity support and fixes advertising restart when multiple identities share the connection pool.

- Add CHIP_BLE_MULTI_IDENTITY_SUPPORT Kconfig option that guards code for handling multiple BT identities.
- Restrict BLE manager and advertising arbiter to the Matter BLE identity (kMatterBleIdentity).
- Replace the advertising restart flag with an atomic flag and retry restart when bt_le_adv_start() fails due to lack of available connection object.

#### Testing

Tested with a second BT identity used by Aliro.

### DIFF
--- a/src/platform/Zephyr/BLEAdvertisingArbiter.cpp
+++ b/src/platform/Zephyr/BLEAdvertisingArbiter.cpp
@@ -41,6 +41,22 @@ const BLEAdvertisingArbiter::Request & ToRequest(const sys_snode_t * node)
     return *static_cast<const BLEAdvertisingArbiter::Request *>(node);
 }
 
+#ifdef CONFIG_CHIP_BLE_MULTI_IDENTITY_SUPPORT
+
+// Check if connection identity matches the identity used for advertising
+bool IsOurIdentity(const bt_conn * conn)
+{
+    VerifyOrReturnValue(conn, false);
+
+    bt_conn_info info{};
+    const int err = bt_conn_get_info(conn, &info);
+    VerifyOrReturnValue(err == 0, false);
+
+    return info.id == sBtId;
+}
+
+#endif // CONFIG_CHIP_BLE_MULTI_IDENTITY_SUPPORT
+
 // Notify application about stopped advertising if the callback has been provided
 void NotifyAdvertisingStopped(const sys_snode_t * node)
 {
@@ -81,7 +97,14 @@ CHIP_ERROR RestartAdvertising()
 }
 
 BT_CONN_CB_DEFINE(conn_callbacks) = {
-    .disconnected = [](struct bt_conn * conn, uint8_t reason) { sWasDisconnection = true; },
+    .disconnected =
+        [](struct bt_conn * conn, uint8_t reason) {
+#ifdef CONFIG_CHIP_BLE_MULTI_IDENTITY_SUPPORT
+            // Ignore disconnections from other identities
+            VerifyOrReturn(IsOurIdentity(conn));
+#endif // CONFIG_CHIP_BLE_MULTI_IDENTITY_SUPPORT
+            sWasDisconnection = true;
+        },
     .recycled =
         []() {
             // In this callback the connection object was returned to the pool and we can try to re-start connectable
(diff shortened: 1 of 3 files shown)


markaj-nordic (approved):

src/platform/Zephyr/BLEAdvertisingArbiter.cpp
```diff
@@ -81,24 +97,38 @@ CHIP_ERROR RestartAdvertising()
...
                     {
                         // Starting from Zephyr 4.0 Automatic advertiser resumption is deprecated,
                         // so the BLE Advertising Arbiter has to take over the responsibility of restarting the advertiser.
                         // Restart advertising in this callback if there are pending requests after the connection is released.
-                        RestartAdvertising();
+                        const CHIP_ERROR result = RestartAdvertising();
+                        if (result != CHIP_NO_ERROR)
+                        {
+                            atomic_set(&sRestart, 1);
+                        }
```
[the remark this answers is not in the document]
adigie: I think we can keep it as is to try again even if it fails for other reason.

src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -134,6 +134,21 @@ constexpr uint8_t kMatterBleIdentity = 1;
 constexpr uint8_t kMatterBleIdentity = 0;
 #endif // CONFIG_BT_BONDABLE
 
+#ifdef CONFIG_CHIP_BLE_MULTI_IDENTITY_SUPPORT
+
+bool IsMatterIdentity(const bt_conn * conn)
```
Damian-Nordic, line 139: Consider defining it always, but simply return true for `!CONFIG_CHIP_BLE_MULTI_IDENTITY_SUPPORT`. This will reduce the number of `ifdef`s in the code.

src/platform/Zephyr/BLEAdvertisingArbiter.cpp
```diff
@@ -81,24 +97,38 @@ CHIP_ERROR RestartAdvertising()
...
 
 BT_CONN_CB_DEFINE(conn_callbacks) = {
-    .disconnected = [](struct bt_conn * conn, uint8_t reason) { sWasDisconnection = true; },
+    .disconnected =
+        [](struct bt_conn * conn, uint8_t reason) {
+#ifdef CONFIG_CHIP_BLE_MULTI_IDENTITY_SUPPORT
+            // Ignore disconnections from other identities
+            VerifyOrReturn(IsOurIdentity(conn));
+#endif // CONFIG_CHIP_BLE_MULTI_IDENTITY_SUPPORT
+            atomic_set(&sRestart, 1);
```
Damian-Nordic, line 106: This flag is only used from `disconnected` and `recycled` callbacks. Aren't they executed from the same thread? In which case atomic wouldn't be needed?

Damian-Nordic (approved):

src/platform/Zephyr/BLEAdvertisingArbiter.cpp
adigie: The second `atomic_set` is called from a lambda that runs on a Matter thread.

Damian-Nordic: ah, right.

src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -988,6 +1023,10 @@ void BLEManagerImpl::HandleDisconnect(struct bt_conn * conId, uint8_t reason)
 ssize_t BLEManagerImpl::HandleC3Read(struct bt_conn * conId, const struct bt_gatt_attr * attr, void * buf, uint16_t len,
                                      uint16_t offset)
 {
+#ifdef CONFIG_CHIP_BLE_MULTI_IDENTITY_SUPPORT
+    VerifyOrReturnValue(IsMatterIdentity(conId), 0);
+#endif // CONFIG_CHIP_BLE_MULTI_IDENTITY_SUPPORT
```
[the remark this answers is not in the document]
adigie: Let's keep `0` as in line 1034.

ArekBalysNordic (approved):

src/platform/Zephyr/BLEManagerImpl.cpp
adigie: I think that could be confusing, maybe it would be better to define something like `ShouldProcessConnection` and use `IsMatterIdentity` inside. Let's keep it as is for now.

